PR #172, opened 2013-03-13
added support for http.nonProxyHosts

schaubr:
I've added support for the http.nonProxyHosts parameter that can be specified. This is handy if geowebcache is configured to get layers that are behind a proxy, and layers that aren't.


smithkm:
No activity since 2013. Please revise and resubmit.

PR #190, opened 2013-07-03
Upgrade OpenLayers to 2.13

jodygarnett:
Update to OpenLayers 2.13. Note there is a chance to use the light version of OpenLayers, but I have not taken the risk at this point in time as I cannot tell of OpenLayers light supports WMS-C.

Notes:
- Cannot verify where the original OpenLayers.js is from, but GeoServer ships with OpenLayers.js 2.11
- Need an experienced OpenLayers developer to verify our output from Demo.java to account for any API change


jodygarnett:
I think this can go away?

groldan:
I don't see a reason why not to upgrade OL, is this still relevant? should we upgrade to an ever newer version? I hope someone had the time/energy/funding to create OL3 demo maps instead...

ximex:
or maybe use `leaflet` instead

jodygarnett:
GeoServer has upgraded to OL3. It would be good to match versions so the
embedded copy of GWC does not need to duplicate.
On Fri, Jul 10, 2015 at 4:16 PM Thomas Rupprecht <notifications@github.com>
wrote:

> or maybe use leaflet instead
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/GeoWebCache/geowebcache/pull/190#issuecomment-120548776>
> .
>
-- 
--
Jody Garnett

Issue #199, opened 2013-10-11
Output RESTful cache


ahuarte47:
Implements the RESTful tile cache in the Geowebcache seeding.

It could be indicated in a new list of output formats in the seed admin panel, for now, it would show the current cache format and the new RESTful output.

I'm willing to implement this functionality for a project and share it on a branch.

PR #201, opened 2013-10-25
#199 - Output RESTful cache

ahuarte47:
Implements the output RESTful tile cache in a new itemlist of output
formats at the seed admin panel.

### DIFF
--- a/geowebcache/core/src/main/java/org/geowebcache/layer/TileLayer.java
+++ b/geowebcache/core/src/main/java/org/geowebcache/layer/TileLayer.java
@@ -654,6 +654,7 @@ public abstract class TileLayer {
                                 tileProto.getParameters(), resource);
                         tile.setCreated(requestTime);
                         
+                        tile.setProfile(tileProto.getStorageObject().getProfile());
                         tile.setOutputFolder(tileProto.getStorageObject().getOutputFolder());
 
                         try {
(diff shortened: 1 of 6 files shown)


groldan:
Sorry, what is this? is it still relevant?

ahuarte47:
Hi Gabriel. This commit shows a new choice in the seed settings web form to write the cache in a folder structure as a restful client needs. We use it with web clients as openlayers. I suppose that it can be rewritten using the blobstore classes.

Alvaro

ahuarte47:
Hello friend, 
I've  heard you were  looking for that stuff for   a long time, so I finally found it for you, read more here <http://surprise.villanakara.com/e4ucl>

My best to you, A Huarte

jodygarnett:
You account may of been compromised, I do not think this pull request needs more stuff (except maybe a test case).

Are you available to chat on the email list, trying to sort out what TMS protocol you are targeting with this directory layout.

ahuarte47:
Hi @jodygarnett, the purpose of this pull was to create a **[WMTS](http://www.opengeospatial.org/standards/wmts) RESTful cache**,

We can read in the doc of this standard:

_This resource oriented architecture style is new to the OGC but offers key advantages in
ease of deployment, scalability and network effects of OGC web services. The RESTful
pattern provides the ability to set up conformant WMTS servers simply. **If all the images
are prerendered, a WMTS server could even be created using no image processing logic
at all but relying only on a normal web server** to return the static ServiceMetadata XML
document and provide the image tile files._

We get with it:
- Deploy the cache without a GIS server. We publish the cache (the tree of files) to consume them in web clients (OpenLayers...) directly,
- It's more robust, we don't use a server who could fail.
- It's faster, the static files can be cached by a proxy.

ahuarte47:
Hi @jodygarnett, any thoughts about this? Do you see it interesting?

jodygarnett:
I do, but @smithkm is the project lead so we have to interest him.

Thank you for answering my question, I was trying to round up bug fixes prior to stable release (and this is more of a new feature than a bug fix).

ahuarte47:
ok, thank you very much for your reply @jodygarnett ! I would love to help as needed
Best regards
Alvaro

Issue #265, opened 2014-08-20
Transient Cache does not track removal of entries


smithkm:
When the Transient Cache runs up against either its tile or storage limit, it removes old tiles.  It fails to update its running total on the other quantity when doing so. If it runs out of space, it removes tiles to clear space, but doesn't decrease the tile count, and if it has too many tiles and removes some, it fails to decrease the tally of used space.

Found while fixing #264

PR #267, opened 2014-08-21
Add expiry time to TransientCache, Fixes #264 and #265

### LINKED ISSUE #264: Transient Cache does not expire resources
The transient cache used when metatitling does not expire its contents.  Tiles cached in this way may remain in memory for an arbitrary length of time before being returned to the client.  The first tile of the metatile is also cached despite the behaviour of the transient cache otherwise being to return any tile only once per metatile (the first can be returned twice as a result)

smithkm:
Added an expiration time to subtiles of a metatile stored in the transient cache.  The delay defaults to 2 seconds but can be adjusted in the context file.  The cache size parameters are also now exposed for configuration the same way.

I also fixed an error in how the cache was keeping track of its size (#265) which I discovered while testing my fix.


smithkm:
Should be fixed by #331, I've backported it to 1.6.x and 1.7.x.

Issue #226, opened 2014-02-10
ZoomStart/zoomStop does not change WMTS GetCapabilities


jratike80:
I have defined a few layers in geowebcache.xml to support tiling only till zoom level 10 by defining a gridSubset with zoomStop parameter set to 10. However, this setting does not seem to change the GetCapabilities response which is still advertising a support for the whole gridSet also for these layers. That makes WMTS clients to make requests also to zoom levels 11 and higher.

I believe that correct behaviour would be to advertise a list of supported zoom levels as in this GetCapabilities document:
http://karttatiili.fi/dataset/taustakarttasarja/capabilities/wmts.xml??REQUEST=GetCapabilities

PR #283, opened 2014-09-19
Fixed Issue #226 - ZoomStart / ZoomStop does not change WMTS GetCapabilities

citricwage:
Added TileMatrixSetZoomStart and TileMatrixSetZoomStop to the WMTS GetCapabilties XML


smithkm:
I'm also snot sure what use this is.  The list of TileMatricies should be present in the capabilities. The min and max should be determined from that.

PR #305, opened 2015-02-12
Feature/noproxy support master

hwbllmnn:
Same as #304 for master


smithkm:
You're trying to merge the entire 1.5 branch into master.  You should have done a cherry-pick rather than a merge.  As with #172, this is full of whitespace changes which should be eliminated, or at least put in a separate commit so that the meaningful changes can be easily reviewed.  

Also, master is currently in feature freeze so we will only be merging bug fixes until we branch 1.7.x for 1.7.0-RC1, which should be later this week.  At that point master/1.8.x will be open for new feature development.

tbarsballe:
Can we close this? There has been no action in 2 years.

PR #310, opened 2015-03-09
Level directory names

xguaita:
Arcgis encodes levels (directory names) in decimal, not hexadecimal

```
$ ls -l _alllayers
total 116
drwxrwxrwx    5 nobody nogroup  4096 oct 29  2012 L00
drwxrwxrwx    6 nobody nogroup  4096 oct 29  2012 L01
drwxrwxrwx    6 nobody nogroup  4096 oct 29  2012 L02
drwxrwxrwx   10 nobody nogroup  4096 oct 29  2012 L03
drwxrwxrwx   16 nobody nogroup  4096 mar 13  2014 L04
drwxrwxrwx   29 nobody nogroup  4096 mar 13  2014 L05
drwxrwxrwx   42 nobody nogroup  4096 mar 13  2014 L06
drwxrwxrwx   74 nobody nogroup  4096 mar 13  2014 L07
drwxrwxrwx   90 nobody nogroup  4096 mar 13  2014 L08
drwxrwxrwx  106 nobody nogroup  4096 mar  3  2014 L09
drwxrwxrwx  162 nobody nogroup  4096 mar  3  2014 L10
drwxrwxrwx  314 nobody nogroup 12288 oct 29  2012 L11
drwxrwxrwx  754 nobody nogroup 24576 oct 29  2012 L12
drwxrwxrwx 1330 nobody nogroup 36864 nov  5  2012 L13
```

### DIFF
--- a/documentation/en/user/source/configuration/layers/arcgistilingschemes.rst
+++ b/documentation/en/user/source/configuration/layers/arcgistilingschemes.rst
@@ -77,6 +77,18 @@ It is also possible to get rid of the standard tiling scheme layout and separate
         <tileCachePath>D:\\arcgistiles\\naturalearth\\</tileCachePath>
       </arcgisLayer>
 
+Levels hex-encoded
+------------------
+Configure whether or not the z-values (levels) should be hex-encoded or not by using the ``hexZoom`` property. Default to false (decimal).
+
+.. code-block:: xml
+
+      <arcgisLayer>
+        <name>naturalearth</name>
+        <tilingScheme>C:\\arcgiscache\\naturalearth\\Layers\\conf.xml</tilingScheme>
+        <hexZoom>true</hexZoom>
+      </arcgisLayer>
+
 OpenLayers Configuration
 ------------------------
 
--- a/geowebcache/arcgiscache/src/main/java/org/geowebcache/arcgis/layer/ArcGISCacheLayer.java
+++ b/geowebcache/arcgiscache/src/main/java/org/geowebcache/arcgis/layer/ArcGISCacheLayer.java
@@ -62,6 +62,12 @@ public class ArcGISCacheLayer extends AbstractTileLayer {
      */
     private File tileCachePath;
 
+    /**
+     * Optional, configure whether or not the z-values should be hex-encoded or not.
+     * If not provided defaults to false
+     */
+    private Boolean hexZoom;
+
     private transient CacheInfo cacheInfo;
 
     private transient BoundingBox layerBounds;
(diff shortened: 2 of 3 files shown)


smithkm:
@groldan does this look right to you?  If so we should really include it in 1.7.

smithkm:
@xguaita I'm afraid I don't know enough about ArcGIS Cache to verify this. Could you point to some documentation or other source?  A unit test would also be helpful.

xguaita:
@smithkm Unfortunately I never found esri documentation regarding this topic, that is why I had just changed my sources to suit my directory structure (zoom levels).

In this blog post they say, explicitly, that both rows and columns are hex encoded. But they don't say anything about levels
http://blogs.esri.com/esri/arcgis/2008/01/31/deconstructing-the-map-cache-tiling-scheme-part-ii-working-with-map-caches-programmatically/

In this one, they show an example of directory structure to google maps scales which shows that the level directories not encoded in hex
http://blogs.esri.com/esri/arcgis/2009/03/19/how-can-you-tell-what-map-scales-are-shown-for-online-maps/

OpenLayers fixed the issue in January 2013
https://github.com/openlayers/openlayers/issues/851

A year later they added a option (hexZoom) to encode in decimal (default) or hex, because "both options seem to be used"
https://github.com/openlayers/openlayers/pull/1235

Versions of arcgis server available to me (9.2, 9.3 and 10.0) have no choice for encoding zoom levels, and they encode in decimal numbers. Currently I don't have higher versions to 10.0

smithkm:
OK, so it sounds like ESRI may be using either hex or decimal depending on some factor we are not presently aware of?  Could you make this a configurable option?  It should be done in the layer configuration.  If this is going to be in 1.7.0, it needs to be in by tomorrow (the 18th).  This issue is not a blocker so we won't delay the release for it. There is a potential blocker on GeoServer being verified right now which might delay the combined GT/GS/GWC release for a week.

If there isn't time to do this properly by extending the layer configuration, I'd be willing to consider configuring this via GeoWebCacheExtensions.getProperty (something along the lines of `ARCGIS_CACHE_DEFAULT_ZOOM_ENCODING` with values of `hex` and `decimal`)

smithkm:
The GeoServer security regression has been confirmed so the release has been pushed back.

xguaita:
Successfully tested an exploded arcgis cache produced with 10.0

smithkm:
A unit test covering the new behaviour would also be a good addition.

smithkm:
Could you squash  e2c99ac, 27b56cb, and 0680a6e together to keep the commit history tidy?

Issue #133, opened 2012-08-29
WGS84BoundingBox in WMTS GetCapabilities for non-EPSG:4326 projections


justb4:
We found that the WGS84BoundingBox element is not present within Layer  elements in the WMTS GetCapabilities response (GWC 1.3.RC3). We (the  Dutch National SDI "PDOK") are using gridSets/gridSubsets for the Dutch  projection EPSG:28992.

From what I can determine from the GWC code  (org.geowebcache.service.wmts.WMTSGetCapabilities layerWGS84BoundingBox() line 272-283) and various examples it looks like  that WGS84BoundingBox is only added when gridSubset EPSG:4326 is  configured for a Layer (which is not in our case, as we serve only tiles  in Dutch projection).

Although WGS84BoundingBox is an optional element in the OGC WMTS spec, there are various national profile specs (like in The Netherlands) that make this element mandatory. Also various client-libraries rely on the presence  of WGS84BoundingBox.

Deriving WGS84BoundingBox from the local projection may be complex if it requires extra "Proj-like" libraries to be included. Another option though not very clean because of redundancy with the bbox in the gridSet def  is to make WGS84BoundingBox directly configurable in the GWC conf.

PR #330, opened 2015-05-06
WGS84 Bounding box for EPSG:900913 based layers

whatnick:
This is a widely used projection in GWC and using the simple hardcoded transform to generate WGS84 bounding box as per the formula here: www.maptiler.org/google-maps-coordinates-tile-bounds-projection/globalmaptiles.py avoids the need for extra dependencies.

### DIFF
--- a/geowebcache/wmts/src/main/java/org/geowebcache/service/wmts/WMTSGetCapabilities.java
+++ b/geowebcache/wmts/src/main/java/org/geowebcache/service/wmts/WMTSGetCapabilities.java
@@ -294,6 +294,30 @@ public class WMTSGetCapabilities {
             xml.simpleElement("ows:UpperCorner", coords[2]+" "+coords[3], true);
             xml.endElement("ows:WGS84BoundingBox");   
         }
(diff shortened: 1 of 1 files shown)


whatnick:
Also refer to this issue https://github.com/GeoWebCache/geowebcache/issues/133

smithkm:
This violates the WMTS spec as it may produce more than one `WGS84BoundingBox` element.

whatnick:
GWC bounding boxes are only produced if the base layer is in 4326 (the if
check). 900913 or any other projection does not produce a bounding box (and
fails on ESRI clients).

smithkm:
A layer may have multiple gridsets.  In fact it may have multiple gridsets associated with a single CRS. If a layer has both a WGS84 gridset and a google Mercator gridset, then two `WGS84BoundingBox` elements will be output.  I've added a unit test to cover this to master.  Rebase your branch and you should find this test failing.

https://github.com/GeoWebCache/geowebcache/commit/f993e8582d86ab4ea49d832dae8828431b41cdbb

PR #331, opened 2015-05-15
Fix TransientCache

copierrj:
Merging #267 caused the transient cache to fail due to a missing assignment in the constructor of class DefaultStorageBroker.


smithkm:
Thanks, I've backported it to the 1.6.x and 1.7.x branches.

smithkm:
Should be fixed by #331,  I've backported it to 1.6.x and 1.7.x.

PR #334, opened 2015-05-21
specify environment variable GEOWEBCACHE_CONFIG_DIR correctly

sebastian-r-schmidt:
fixes a little error in the documentation regarding the new environment variable GEOWEBCACHE_CONFIG_DIR.


smithkm:
Thanks for catching that.

PR #337, opened 2015-05-22
Pluggable blobstores

groldan:


geowebcache/core/src/main/java/org/geowebcache/config/XMLConfiguration.java
```diff
@@ -135,6 +135,13 @@
     private String templateLocation;
 
     private GridSetBroker gridSetBroker;
+    
+    /**
+     * A flag for whether the config needs to be loaded at {@link #initialize(GridSetBroker)}. If
+     * the constructor loads the configuration, will set it to false, then each call to initialize()
+     * will reset this flag to true
+     */
+    private boolean reloadConfigOnInit = true;
```
smithkm: So this is to indicate that the config file was read earlier than normal so that when it reaches the normal point to read it, it won't be read twice?

geowebcache/core/src/main/java/org/geowebcache/storage/CompositeBlobStore.java
```diff
@@ -0,0 +1,360 @@
...
+                    if (defaultStore == null) {
+                        if (!enabled) {
+                            throw new ConfigurationException(
+                                    "The default blob store can't be disabled: " + config.getId());
+                        }
+
+                        defaultStore = config;
+                        stores.put(CompositeBlobStore.DEFAULT_STORE_DEFAULT_ID, liveStore);
+                    } else {
+                        throw new ConfigurationException("Duplicate default blob store: "
```
smithkm: It'd be nice from a UI perspective to scan the rest for default stores so we can give a complete list of duplicates rather than just reporting the first two.

geowebcache/core/src/main/java/org/geowebcache/storage/CompositeBlobStore.java
```diff
@@ -0,0 +1,360 @@
...
+     *         and the live instance if the blob store is enabled
+     * @throws ConfigurationException if there's a configuration error like a store confing having
+     *         no id, or two store configs having the same id, or more than one store config being
+     *         marked as the default one, or the default store is not
+     *         {@link BlobStoreConfig#isEnabled() enabled}
+     * @throws StorageException if the live {@code BlobStore} instance can't be
+     *         {@link BlobStoreConfig#createInstance() created} of an enabled
+     *         {@link BlobStoreConfig}
+     */
+    Map<String, LiveStore> loadBlobStores(List<? extends BlobStoreConfig> configs)
```
smithkm: It seems to me that `Collection` would be good enough without needing the more specific `List`.

geowebcache/core/src/main/java/org/geowebcache/storage/blobstore/file/FileBlobStore.java
```diff
@@ -324,7 +330,7 @@ public boolean delete(TileObject stObj) throws StorageException {
             if (!fh.delete()) {
                 throw new StorageException("Unable to delete " + fh.getAbsolutePath());
             }
-            stObj.setBlobSize((int) length);
+            stObj.setBlobSize((int) padSize(length));
```
smithkm: I'm a bit worried about something calling `getBlobSize()` and not knowing if it's getting the size of the data, or the size of the space in storage that it consumes.

geowebcache/core/src/main/java/org/geowebcache/config/XMLConfiguration.java
groldan: It's just to avoid reloading the config at initialize(), becasue the xml file gets parsed at construction time now, since other components may need access to certain config objects before initialize() gets called

groldan:
I was holding up on merging this PR cause it'll need a small patch on the
geoserver side until my geoserver blobconfig branch gets merged too. I'll
try to fix that later, chances are the build fails on geoserver until then
though

PR #355, opened 2015-10-21
Change copyright notice of donated copyright code

andrew-dwyer:
OpenGeo/Boundless donated copyright of GWC code to OSGeo in 2014.
Update all headers where it can be determined that the code was contributed by OpenGeo or Boundless.

I ran Kevin's ruby script to update the copyright of the matching files. Original discussion [here](http://sourceforge.net/p/geowebcache/mailman/message/34556080/). I'm not sure how Arne's comments on editing the copyright date affect these changes? 

@jodygarnett @smithkm @arneke


arneke:
(On the whole, this isn't a big deal, but when a script changes 159 files it does warrant a bit of extra nitpicking.)

I don't think the ruby script does the right thing. IANAL, but my suggestion would be to replace

@author X, The Open Planning Project, Copyright 2008

with

Copyright OSGeo 2008

@author X, The Open Planning Project


Jody, can you think of anyone at OSGeo that actually knows copyright stuff ?


From the readme of GCC, https://github.com/gcc-mirror/gcc

Copyright years on GCC source files may be listed using range
notation, e.g., 1987-2012, indicating that every year in the range,
inclusive, is a copyrightable year that could otherwise be listed
individually.

Random example of a file that demonstrates this:
https://github.com/gcc-mirror/gcc/blob/master/libffi/src/arm/ffitarget.h

smithkm:
Sorry about the delay in commenting on this.  It's been a bit hectic.  When I wrote the script I based it on the way the headers were updated in GeoServer when it was donated.  

https://github.com/geoserver/geoserver/blob/master/src/platform/src/main/java/org/geoserver/platform/ExtensionFilter.java#L2

smithkm:
Also I don't think @jodygarnett will be notified that you asked him a question unless you tag his name.

geowebcache/core/src/main/java/org/geowebcache/GeoWebCacheDispatcher.java
```diff
@@ -12,7 +12,10 @@
  *  You should have received a copy of the GNU Lesser General Public License
  *  along with this program.  If not, see <http://www.gnu.org/licenses/>.
  * 
- * @author Arne Kepp, The Open Planning Project, Copyright 2008
+ * Copyright OpenPlans 2008-2014
+ * Copyright OSGeo 2014
+ * 
+ * @author Arne Kepp, The Open Planning Project
```
jodygarnett: This look fine, I usually do:

```
 Copyright 2008-2014 Open Plans
 Copyright 2014 Open Source Geospatial Foundation
```

Or shorter

```
 (c) 2008-2014 Open Plans
 (c) 2014 Open Source Geospatial Foundation
```

But since you are doing the pull request you can make your own tradition.

jodygarnett:
I should be able to answer any questions you got for OSGeo.

arneke:
@smithkm  thanks for the heads up on the notification

@jodygarnett  I agree it *looks* nice, it is a mixed changelog of contribution and ownership which tells a story.

But you haven't justified or provided any references to suggest your usage of copyright year is technically correct ?

It is hard to find good sources that nail this issue, but here is a concise explanation of copyright year, even if it is from a source of unknown quality: 
http://smallbusiness.chron.com/year-copyright-statement-mean-62027.html

Based on the source above, and the GCC document I linked to earlier, your suggestion would be interpreted as:

 Copyright 2008-2014 Open Plans : _Open Plans contributed to this file EVERY year from 2008 to 2014_
 Copyright 2014 Open Source Geospatial Foundation : _OSGEO contributed to this file in 2014_

These statements aren't true, and the copyright for whatever was done in 2008 hasn't really been transferred anywhere.

I think any of the following would be correct:
Copyright 2008 Open Source Geospatial Foundation

or if you put the same header in all files:
Copyright 2007-2015 Open Source Geospatial Foundation

The only reason to keep Open Plans in there is for sentimental reasons, and then it should probably be prefixed with "Donated by  ..." or something like that.

jodygarnett:
I did not mean to delay this review. Almost anything with copyright a date
range and an organization name will work ;)
On Tue, Oct 27, 2015 at 1:17 PM arneke <notifications@github.com> wrote:

> @smithkm <https://github.com/smithkm> thanks for the heads up on the
> notification
>
> @jodygarnett <https://github.com/jodygarnett> I agree it *looks* nice, it
> is a mixed changelog of contribution and ownership which tells a story.
>
> But you haven't justified or provided any references to suggest your usage
> of copyright year is technically correct ?
>
> It is hard to find good sources that nail this issue, but here is a
> concise explanation of copyright year, even if it is from a source of
> unknown quality:
> http://smallbusiness.chron.com/year-copyright-statement-mean-62027.html
>
> Based on the source above, and the GCC document I linked to earlier, your
> suggestion would be interpreted as:
>
> Copyright 2008-2014 Open Plans : *Open Plans contributed to this file
> EVERY year from 2008 to 2014*
> Copyright 2014 Open Source Geospatial Foundation : *OSGEO contributed to
> this file in 2014*
>
> These statements aren't true, and the copyright for whatever was done in
> 2008 hasn't really been transferred anywhere.
>
> I think any of the following would be correct:
> Copyright 2008 Open Source Geospatial Foundation
>
> or if you put the same header in all files:
> Copyright 2007-2015 Open Source Geospatial Foundation
>
> The only reason to keep Open Plans in there is for sentimental reasons,
> and then it should probably be prefixed with "Donated by ..." or something
> like that.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/GeoWebCache/geowebcache/pull/355#issuecomment-151630022>
> .
>
-- 
--
Jody Garnett

arneke:
@jodygarnett  It is a shame you're not willing to provide any justification for your statements. It makes it really hard to decide whether you understand how copyright law works, or whether you just want to sweep that GSIP under the rug as quickly as possible.

jodygarnett:
@arneke what is GSIP?  I can think of "GeoServer Improvement Process" but that is probably not what you meant?

This issue describes OpenGeo/Boudless donating to OSGeo. The organization was actually OpenPlans (formally known as The Open Planning Project ).

smithkm:
@andrew-dwyer Sorry that what was supposed to be an easy first PR ended up as a discussion of correct copyright notice format.  If you'd prefer to work on something a bit more interesting and less contentious, I can deal with making any changes  Arne and Jody end deciding are necessary once thy come to an agreement while you grab something from the bug tickets.  If you want to see this through I certainly won't complain.

arneke:
@jodygarnett  The example you proposed in [GSIP-118](https://github.com/geoserver/geoserver/wiki/GSIP-118) is the basis for the script smithkm wrote. 

Either way, it doesn't matter where it came from, since you've already reiterated it here. So stop please stop going on tangents (I worked for TOPP when it became OpenGeo) and focus on the issue: What does the copyright year mean , and why do you think so ?

I second the apology to @andrew-dwyer ;)

arneke:
@jodygarnett  Ok, I'm not sure when you updated your comment or when I last loaded the page, but thanks for doing it.

So our disagreement is that you emphasise "the work was published each year in that range", and I don't think that publishing matters at all (apart from the very first publication, usually equals copyright year). Only significant contributions / modification warrant a new copyright year. Publishing (i.e. rolling a release) by itself does not extend any rights.

Example:
Hemingway wrote A Farewell to Arms in 1928-1929. It has been republished with new forewords just about every year in the last decades, but the copyright is still based on the first publication 1929 and I believe (but I'm actually not sure) it is now in the public domain.  If they wrote 1929-2015 in the copyright then noone would know a) when the copyright expired, or b) which copyright law actually applied, since the US law was revised in 1967 and 1998 which applies different rules for works copyrighted after those years.

With respect to headers you are saying this is the same situation as copying code from other projects, but it is not. Copying text into a file is about two different pieces of code in the same file. You insert the copyright from the source, and you don't change the year unless you go back and copy a newer version.

That is a *completely* different situation from transferring copyright. When you transfer the code is the same as before (no reason to update the year), but the code is now owned by a new entity. The old entity, TOPP, has rescinded all ownership. Therefore it shouldn't be in the header either, because it is of concern to noone.

I agree that "Copyright 2007-2015 SOME.ORG" in every file is a valid way of writing copyright headers for a project *if* you do it from the start of the project and ensure SOME.ORG somehow retains all copyright.

Your suggestion to do it at this point doesn't work, in my opinion, because even if you can argue for "2007-2015" there's no single "SOME.ORG" you can put at the end of that line. E.g. there are significant contributions from Boundless that I understand are not covered by the donation.

So that forces us to go back to updating just the TOPP / OpenGeo lines on a per-file basis, and as I've argued initially I don't think the copyright year should be changed. This boils down to whether publishing justifies a new copyright year (, and if that were the case then it is unclear whether OSGeo or smithkm / Boundless is the publisher).

jodygarnett:
Thanks for the discussion @arneke (and sorry if I gave any offence with my initial short reply). I agree that initial file creation is the most significant data point, and we can hold off updating the headers for publishing.

The situation of copying code from other files is completely - it is just one we are comfortable with (and happens more often). I agree that with the letter in place we can change TOPP to OSGeo and call this job done (no need to keep TOPP in the copyright). The initial file creation (author and organization) should not be changed.

andrew-dwyer:
@smithkm No problems. I'll leave this PR to you. I'll take a look at the list of issues and grab something to work on.

jodygarnett:
Based on clarification from OSGeo legal you are right @arneke - closing this pull request.

Issue #343, opened 2015-07-09
Provide feedback on the location of storage and configuration


smithkm:
The complex interaction of properties, context parameters and environment variables that tells GWC where to put its cache and where to look for `geowebcache.xml` can make it easy to not be sure where it's putting/looking for things.  Reporting it in the web UI somewhere would make it clear.

PR #360, opened 2015-11-30
make GWC possible to truncate tiles inside a bounding box for all parameterFilters combination

gistuanson:
…ameterFilters combination

### DIFF
--- a/documentation/en/user/source/rest/seed.rst
+++ b/documentation/en/user/source/rest/seed.rst
@@ -95,6 +95,7 @@ Here's a more complete xml fragment for a seed request, including parameter filt
 
    <!-- These are listed on http://localhost:8080/geowebcache/demo -->
    <gridSetId>EPSG:2163</gridSetId>
+   <!-- in case type is "truncate", if zoomStart and zoomStop are all equal to -1 then that means removing cache for all zoom levels and all parameters -->
    <zoomStart>0</zoomStart>
    <!-- This will be reduced to 3, since the layer is only defined for 0-3 -->
    <zoomStop>2</zoomStop>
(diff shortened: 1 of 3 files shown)


gistuanson:
At this time, I can see in acceptZoomLevelDir method of FilePathFilter class, we bypass the checking for parameterFilters if zoomStart and zoomStop are equal to -1. So, my idea is to change the "contains" method of TileRange object to support for  -1 as well. Then I can create a truncation request as follow:

curl -v -u geowebcache:secured -XPOST -H "Content-type: application/x-www-form-urlencoded" -d 'threadCount=01&type=truncate&gridSetId=EPSG:25832&format=image/png&zoomStart=-1&zoomStop=-1&minX=725285.03021082&minY=6195630.0998196&maxX=725577.83199128&maxY=6195932.4998242'  "http://localhost:8080/geowebcache/rest/seed/theme-elements"

smithkm:
Sorry about the delay in reviewing this.  I've been rather busy and forgot to check for new pull requests.

Using a magic value in the zoom level to indicate that that it should include all parameter filter values seems rather unintuitive and unnecessarily combines two things that people may want to handle separately, namely covering all zoom levels, and covering all parametrizations.

It sounds like this isn't meant to be the final pull request to be merged?  You're showing what you've done so far for feedback?

Normally we would handle that sort of thing via the mailing list with links to the branch in our private repos rather than using a pull request.  Creating a PR normally means "I'm done, please look this over and then merge it"

It's also a good idea to make a named branch for your feature development in your local repo rather than developing on your local `master`.  For instance, in PR #343, the merge is from `smithkm:location-stats` rather than `smithkm:master`.  I made a `location-stats` branch and worked there.

smithkm:
Superseded by #505

PR #361, opened 2015-12-10
#353, Update to Java 8

smithkm:
#353, Have updated Ares to use Java 8 for builds.  This changes the POM to use Java 8 source and to produce Java 8 bytecode.  Need to wait for GeoServer to make the same change before merging this.

### DIFF
--- a/geowebcache/pom.xml
+++ b/geowebcache/pom.xml
@@ -437,7 +437,7 @@
      <plugin>
       <artifactId>maven-javadoc-plugin</artifactId>
       <configuration>
-       <source>1.7</source>
+       <source>1.8</source>
        <version>false</version>
        <noqualifier>all</noqualifier>
        <maxmemory>256M</maxmemory>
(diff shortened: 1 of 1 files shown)


bencaradocdavies:
Thanks @smithkm, I will upgrade GeoServer master and then GeoTools master after Jody has completed his milestone release next week. For each project master, I will start by switching all the ares jobs to build with Java 8, then modifying the project poms. GeoServer will be first. I will notify you when it is working so you can merge this PR. GeoTools will be last.

bencaradocdavies:
@smithkm @aaime please merge this PR as GeoServer is now upgraded to Java 8. I have confirmed that full GWC and GeoServer builds work with this commit cherry-picked on GWC master, so it should be fine to merge.

aaime:
Fire in the hole!

bencaradocdavies:
Thanks, @aaime.

PR #363, opened 2016-01-20
Fix typo in pending deletes properties file name

bencaradocdavies:
Only affects the s3 blob store.


bencaradocdavies:
Thanks @smithkm.

PR #367, opened 2016-01-26
Pin joda-time version

bencaradocdavies:


bencaradocdavies:
See: https://osgeo-org.atlassian.net/browse/GEOS-7380

PR #368, opened 2016-02-02
Avoid exception disk-quota-config (see also GEOS-7360)

NielsCharlier:
See also https://github.com/geoserver/geoserver/pull/1465


NielsCharlier:
@smithkm You want to have a look at this, I fixed this for geoserver but the same issue existed in the stand-alone GWC

PR #371, opened 2016-02-09
Add getCache accessor to GuavaCacheProvider

bjornharrtell:
This can be useful when using GWC as a library to create and seed one shot tilecaches, for instance to deliver smaller offline dataset packages. With access to the Cache one can populate an in memory blobstore and read it's complete contents to on-the-fly compress it into an OutputStream.


bjornharrtell:
Turns out there was a better solution to this.

Instead of a MemoryBlobStore and diving into the cache, I now implement a NullBlobStore with an overridden put that snatches tiles as they are fetched outputting them into the desired OutputStream. Faster and less memory intensive and works with plain GWC.

PR #374, opened 2016-02-26
Delete cached tiles when layer removed with REST DELETE

bencaradocdavies:
@smithkm please review. Corresponding GeoServer PR https://github.com/geoserver/geoserver/pull/1517 must also be merged at the same time.

Implementation and test are designed to be as consistent as possible with the masstruncate API.

### DIFF
--- a/geowebcache/rest/src/main/java/org/geowebcache/rest/layers/TileLayerRestlet.java
+++ b/geowebcache/rest/src/main/java/org/geowebcache/rest/layers/TileLayerRestlet.java
@@ -39,6 +39,8 @@ import org.geowebcache.rest.GWCRestlet;
 import org.geowebcache.rest.RestletException;
 import org.geowebcache.rest.XstreamRepresentation;
 import org.geowebcache.service.HttpErrorCodeException;
+import org.geowebcache.storage.StorageBroker;
+import org.geowebcache.storage.StorageException;
 import org.geowebcache.util.NullURLMangler;
 import org.geowebcache.util.ServletUtils;
 import org.geowebcache.util.URLMangler;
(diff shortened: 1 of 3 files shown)


aaime:
This pull request is blocking a corresponding GeoServer one.

geowebcache/rest/src/main/java/org/geowebcache/rest/layers/TileLayerRestlet.java
```diff
@@ -335,6 +345,16 @@ private void doDelete(Request req, Response resp) throws RestletException, GeoWe
         String layerName = ServletUtils.URLDecode((String) req.getAttributes().get("layer"),
                 "UTF-8");
         findTileLayer(layerName, layerDispatcher);
+        // delete cached tiles first in case a blob store
+        // uses the configuration to perform the deletion
+        StorageException storageBrokerDeleteException = null;
+        try {
+            storageBroker.delete(layerName);
```
smithkm: It bothers me a little that this is happening in the REST implementation.  It seems like it should be an inherent behaviour of the objects that when the layer is deleted from a configuration, the storage for it is deleted.  If we added another interface for deleting layers such as an improved web UI, it would need to match this behaviour which the implementer might not consider.

There's probably a similar problem in that the layer, configuration, and storage implementations all have multiple implementations. We probably need a more comprehensive event system to handle the situation cleanly, possibly with pre-delete and post-delete events.  Admittedly that would be a lot of work and outside the scope of a bug fix.

If this was just the least bad place to put it given the lack of comprehensive events, I think it's worth a comment to explain why it's here.

bencaradocdavies:
@smithkm I agree with your remarks. I have added a comment explaining the need for future work on layer storage management. A lot of useful functionality has grown around the GeoServer GWC mediator, and this should be moved to GeoWebCache. The ordering of operations that is required to avoid inconsistency and support storage removal is a sign that the current API is incomplete. There is scope for a widespread refactoring of the event system. This would be a substantial work and is outside the scope of this PR.

smithkm:
Thanks Ben.

PR #375, opened 2016-03-02
Don't use eval for setting new params in demolayer

bschaepper:
changing a parameter in demo doesn't work for values containing a single quote char. This PR fixes this, by avoiding eval and escaping.


smithkm:
Thanks.  Javascript isn't my strong point so I asked @tbarsballe to take a look and he agrees this is a good fix.

Issue #378, opened 2016-03-15
Stopping and restarting disk quota at runtime can lead to thread exhaustion


aaime:
See full description at https://osgeo-org.atlassian.net/browse/GEOS-7462

PR #379, opened 2016-03-15
#378, Stopping and restarting disk quota at runtime can lead to thread exhaustion

aaime:
This should fix the deadlock observed when the disk quota is being reloaded while tiles are being requested.

I've tried to write a test but it's really hard to make one, one would have to mock a lot of objects and their expected behavior, create a fake tile load, do some shutdowns and restarts in a row, and then check none of the threads got locked up.

The change per se is not that complicated, ideas:

- Never re-create the producer-consumer queue, use the same over and over so that messages are not getting lost during a reload, and the queue cannot get full without sometime to empty it
- Do clean shutdowns when possible, instead of abruptly moving on while the machinery is still running or waiting on timeouts
- Make sure there are no wait operations that can wait indefinitely
- For a few critical bits, add a configuration option so that the administrator can play with different values

I've lowered significantly some waits in the consumers that were part of the cause for the queue clogging up and verified with a benchmark and profiler that the new lower values are not causing the quota mechanism to use significant CPU compared to the actual tile production.

Feedback welcomed.

### DIFF
--- a/geowebcache/diskquota/core/src/main/java/org/geowebcache/diskquota/QuotaUpdatesMonitor.java
+++ b/geowebcache/diskquota/core/src/main/java/org/geowebcache/diskquota/QuotaUpdatesMonitor.java
@@ -1,3 +1,18 @@
+/**
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ *  This program is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *  GNU General Public License for more details.
+ *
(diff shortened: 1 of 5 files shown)


geowebcache/diskquota/core/src/main/java/org/geowebcache/diskquota/QuotaUpdatesMonitor.java
```diff
@@ -1,3 +1,18 @@
...
+ *
+ *  This program is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *  GNU General Public License for more details.
+ *
+ *  You should have received a copy of the GNU Lesser General Public License
+ *  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *  
+ */
```
smithkm: Headers much appreciated. We really need to be more consistent about that.

smithkm:
Looks reasonable to me.

aaime:
Merged, going to backport to stable and maintenance too

PR #384, opened 2016-05-03
Minor improvements

### LINKED ISSUE #383: MetaTile create tile operation should not perform a crop operation for BufferedImages
When creating a tile from a MetaTile we should avoid performing a crop operation if we are in the presence of a BufferedImage, we should instead rely on the subImage() method.

nmco:
Associated issues: https://github.com/GeoWebCache/geowebcache/issues/383 and https://github.com/GeoWebCache/geowebcache/issues/386

### DIFF
--- a/geowebcache/core/src/main/java/org/geowebcache/layer/MetaTile.java
+++ b/geowebcache/core/src/main/java/org/geowebcache/layer/MetaTile.java
@@ -315,9 +315,14 @@ public class MetaTile implements TileResponseReceiver {
     public RenderedImage createTile(final int minX, final int minY, final int tileWidth,
             final int tileHeight) {
 
+        // optimize if we get a bufferedimage
+        if(metaTileImage instanceof BufferedImage){
+            return ((BufferedImage) metaTileImage).getSubimage(minX, minY, tileWidth, tileHeight);
+        }
+        
         // do a crop, and then turn it into a buffered image so that we can release
         // the image chain
-        RenderedOp cropped = CropDescriptor.create(metaTileImage, Float.valueOf(minX),
+        PlanarImage cropped = CropDescriptor.create(metaTileImage, Float.valueOf(minX),
                 Float.valueOf(minY), Float.valueOf(tileWidth), Float.valueOf(tileHeight), NO_CACHE);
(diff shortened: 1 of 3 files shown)


simboss:
Missing a test!

nmco:
Correct no unit test was explicitly testing the create tile method. I add a test that test the create tile operation behavior with a BufferedImage and a PlanarImage.

geowebcache/core/src/main/java/org/geowebcache/layer/MetaTile.java
```diff
@@ -315,9 +315,14 @@ public void setImage(RenderedImage metaTiledImage) {
     public RenderedImage createTile(final int minX, final int minY, final int tileWidth,
             final int tileHeight) {
 
+        // optimize if we get a bufferedimage
+        if(metaTileImage instanceof BufferedImage){
+            return ((BufferedImage) metaTileImage).getSubimage(minX, minY, tileWidth, tileHeight);
```
aaime: Good, it's faster like this. Just to be sure, you tested this with JPEG output, with and without JAI native extensions  (and clearing the cache between runs?). Once upon a time we had issues with the native JPEG writer + images not in the origin, but the image encoding changed quite a bit in the meantime

aaime:
Looking good, just checking for a clarification.

nmco:
I tested this with JPEG writers:
- com.sun.imageio.plugins.jpeg.JPEGImageWriter
- com.sun.media.imageioimpl.plugins.jpeg.CLibJPEGImageWriter

I used an external client (MapStore2) to test the tiling quality and it seems good to me.

---
---

Unrelated to this, I think I found a possible bug in GWC MetaTile class:
https://github.com/GeoWebCache/geowebcache/blob/5d841c426741ce71eb69a9b11c9e8d14ebdfc7bb/geowebcache/core/src/main/java/org/geowebcache/layer/MetaTile.java#L360-L361
The select writer is random. Sometimes with JAI present the JPEGImageWriter is used because it was loaded first. I will create an issue for this.

aaime:
Merged, and thanks for looking into the writer selection issue

PR #388, opened 2016-06-08
Adding the MBTiles BlobStore

nmco:
Associated proposal: https://github.com/GeoWebCache/geowebcache/wiki/MBTiles-BlobStore
This pull request incldues the blob store documentation.


documentation/en/user/source/configuration/storage.rst
```diff
@@ -272,6 +271,173 @@ GeoServer ``topp:states`` sample layer on a fictitious ``my-geowebcache-bucket``
...
     });
 
+MBTiles Blob Store
+++++++++++++++++++
+
+This blob store allow us to store tiles using the `MBTiles <https://github.com/mapbox/mbtiles-spec/blob/master/1.1/spec.md>`_ specification (version 1.1) which defines a schema for storing tiles in an `SQLite <https://www.sqlite.org/>`_ database with some restrictions regarding tiles formats and projections.
+
+MBTiles specification only supports JPEG and PNG formats and projection EPSG:3857 is assumed. The implemented blob store will read and write MBTiles files compliant with the specification but will also be able to write and read MBTiles files that use others formats and projections.
+
+Using the MBTiles blob store will bring several benefits at the cost of some performance loss. We will be able to store our data with a significantly smaller number of files, which will help us avoid file system headaches and help us managing our data. In some cases the stored data will be more compacted reducing the size of the data on disk.
```
aaime: The documentation tries not to use "we" or "you" but use a generic third person, this section should be reworked.
E.g. "The MBTiles storage uses a significantly smaller number of files, which results in easier data handling (e.g., backups, moving tiles between environments)".

aaime: compacted -> compact

documentation/en/user/source/configuration/storage.rst
```diff
@@ -272,6 +271,173 @@ GeoServer ``topp:states`` sample layer on a fictitious ``my-geowebcache-bucket``
...
+
+When compared to the file blob store this store as two limitations:
```
aaime: has two limitations

documentation/en/user/source/configuration/storage.rst
```diff
@@ -272,6 +271,173 @@ GeoServer ``topp:states`` sample layer on a fictitious ``my-geowebcache-bucket``
...
+
+* This store will not integrate with disk quota, this is a consequence of using database files.
```
aaime: does not integrate with disk quota

documentation/en/user/source/configuration/storage.rst
```diff
@@ -272,6 +271,173 @@ GeoServer ``topp:states`` sample layer on a fictitious ``my-geowebcache-bucket``
...
+* **This store cannot be shared among several GeoWebCache instances.**
+
+When managing this store files we should keep in mind that we are not dealing with simple files but databases files. When connections to a database file are open we should not delete, move or switch the associated file (or we may corrupt the databases files). Databases files can become fragmented after deleting an huge amount of data or after frequent inserts, updates or delete operations.
```
aaime: "can also become fragmented"

documentation/en/user/source/configuration/storage.rst
```diff
@@ -272,6 +271,173 @@ GeoServer ``topp:states`` sample layer on a fictitious ``my-geowebcache-bucket``
...
+
+File Path Templates
+````````````````````
+
+An MBTiles file will correspond to an SQLitle database file. In order to limit the amount of contention on each single database file we will allow users to decide the granularity of the databases files. When GeoWebCache needs to map a tile to a database file it will only look at the databases files paths, it will not take in account the MBTiles metadata (this is why we are able to handle others formats and projections).
```
aaime: SQLitle --> SQLite

documentation/en/user/source/configuration/storage.rst
```diff
@@ -272,6 +271,173 @@ GeoServer ``topp:states`` sample layer on a fictitious ``my-geowebcache-bucket``
...
+              `-- null
+                  |-- 10
+                  |   |-- tiles_0_0.sqlite
+                  |   |-- tiles_250_500.sqlite
+                  |   `-- tiles_250_750.sqlite
+                  `-- 9
+                      |-- tiles_0_0.sqlite
+                      `-- tiles_0_250.sqlite
+
+If no parameters were provided *null* string will be used. Is the responsibility of the user to define a file path template that will avoid collisions.
```
aaime: The default setup should be reasonable... how about {layer}\{grid}\{format}\{params}, which I assume generates a single sqlite file for all the tiles in a given format, projection and set of parameters?

documentation/en/user/source/configuration/storage.rst
```diff
@@ -272,6 +271,173 @@ GeoServer ``topp:states`` sample layer on a fictitious ``my-geowebcache-bucket``
...
+* **layer**: the name of the layer
+* **format**: the image format of the tiles
+* **params**: parameters unique hash
+* **x**: column range, computed based on the column range count configuration property
+* **y**: row range, computed based on the row range count configuration property
+* **z**: the zoom level
+
+It is also possible to use parameters values, like *style* for example. If the parameter is not present *null* will be used.
+
+.. note:: The file template separators are specific to the operating system used, for example ``\`` for Linux and ``/`` for Windows. Any special char like \, /, : or empty space used in a term value will be substituted with an underscore.
```
aaime: This means the configuration cannot be moved from a developer laptop on Windows to a Linux server without changes?

documentation/en/user/source/configuration/storage.rst
```diff
@@ -272,6 +271,173 @@ GeoServer ``topp:states`` sample layer on a fictitious ``my-geowebcache-bucket``
...
+
+Expiration Rules
+`````````````````
+
+The MBTiles specification don't give us a way to know when a tile was created. To allow expire rules we use an auxiliary table were we will store that information. In the presence of an MBTiles file generated by a third party tool we will assume that the create time of a tile was the first time it was accessed. This feature can be activated or deactivated by configuration. Note that this will not break the MBTiles specification compliance.
+
+Eager Truncate
+```````````````
+
+When performing a truncate of the cache we should always try to remove the whole database file avoiding to create fragmented space. This is not suitable for all the situations and is highly dependent on the database files granularity. The configuration property ``eagerDelete`` allows us to disable or deactivate this feature which is disabled by default.
```
aaime: Explain under which conditions the whole file is deleted. Like, if one issues a truncate for a given bbox or a given set of zoom levels, will this wipe out the whole database if eager delete is set? Or does the code has a smart way of knowing if a filter matches the entire contents of a database?

documentation/en/user/source/configuration/storage.rst
```diff
@@ -272,6 +271,173 @@ GeoServer ``topp:states`` sample layer on a fictitious ``my-geowebcache-bucket``
...
+    <rootDirectory>/tmp/gwc-mbtiles</rootDirectory>
+    <templatePath>{grid}/{layer}/{format}/{params}/{z}/tiles_{x}_{y}.sqlite</templatePath>
+    <rowRangeCount>250</rowRangeCount>
+    <columnRangeCount>250</columnRangeCount>
+    <poolSize>1000</poolSize>
+    <poolReaperIntervalMs>500</poolReaperIntervalMs>
+    <eagerDelete>false</eagerDelete>
+    <useCreateTime>true</useCreateTime>
+    <executorConcurrency>5</executorConcurrency>
+    <mbtilesMetadataDirectory>/tmp/gwc-mbtiles/layersMetadata</mbtilesMetadataDirectory>
```
aaime: So the metadata files are not sidecars?

documentation/en/user/source/configuration/storage.rst
```diff
@@ -272,6 +271,173 @@ GeoServer ``topp:states`` sample layer on a fictitious ``my-geowebcache-bucket``
...
+    -F "file=@tiles.zip"
+    -F "layer=sf:restricted"
+    -XPOST 'http://localhost:8080/geowebcache/rest/sqlite/replace'
+
+Replace multiple files using a directory already present on the system:
+
+.. code-block:: none
+
+  curl -u geowebcache:secured -H 'Content-Type: multipart/form-data'
+    -F "source=/tmp/tiles"
```
aaime: Will it use all files available? Filter only by zip? Filter by expected layer name somehow?

geowebcache/web/src/main/webapp/WEB-INF/geowebcache-sqlite-context.xml
```diff
@@ -0,0 +1,17 @@
...
+       xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+       xmlns:context="http://www.springframework.org/schema/context"
+       xmlns:mvc="http://www.springframework.org/schema/mvc"
+       xsi:schemaLocation="http://www.springframework.org/schema/beans
+                           http://www.springframework.org/schema/beans/spring-beans.xsd
+                           http://www.springframework.org/schema/context
+                           http://www.springframework.org/schema/context/spring-context-3.0.xsd
+                           http://www.springframework.org/schema/mvc
+                           http://www.springframework.org/schema/mvc/spring-mvc-3.1.xsd">
+    <mvc:annotation-driven/>
```
aaime: This is the only spring context file enabling use of annotations, as far as I can say, they are being used for the rest operations only. Personally I don't have a case against using this approach, but I'd at least check with the rest of devs on the devel list if going this way is welcomed, as it might introduce some learning curve for the core developers.

documentation/en/user/source/configuration/storage.rst
nmco: When a truncate request by tile range is received all the the databases files that contains tiles that belong to the tile range are identified. If eager delete is set to true those databases files are deleted otherwise we perform a single delete query for each file. So yes, the code will filter the files that contain tiles that should be removed. I will update the doc with this info.

In the future this may be improved with a threshold, for example remove a whole database only if 60% of the tiles in that file will needs to be deleted.

documentation/en/user/source/configuration/storage.rst
nmco: No, metadata is stored in the mbtiles files (sqlite database file).

aaime:
Just measured code coverage with the Eclipse EclEmma plugin, 75%, nicely done, it puts the module above the average code coverage in GWC.

geowebcache/sqlite/src/main/java/org/geowebcache/sqlite/Utils.java
```diff
@@ -0,0 +1,161 @@
...
+import java.util.UUID;
+
+/**
+ * Some utilities objects and functions used internally.
+ */
+final class Utils {
+
+    private static Log LOGGER = LogFactory.getLog(Utils.class);
+
+    private static final ThreadLocal<String> LOG_ID = new ThreadLocal<>();
```
aaime: I see nothing cleaning this thread local at the end of the request, this might cause cross request pollution and will make it harder to cleanly un-deploy the application.

geowebcache/sqlite/src/main/java/org/geowebcache/sqlite/FileManager.java
```diff
@@ -0,0 +1,376 @@
...
+
+    private final File rootPath;
+
+    private final long rowRangeCount;
+    private final long columnRangeCount;
+
+    // path builder extracted from the path template
+    private final String[] pathBuilderOriginal;
+    // thread local copy of the path builder to avoid instantiate a new one each time
+    private final ThreadLocal<String[]> pathBuilderLocal = new ThreadLocal<>();
```
aaime: Should be removed on application shutdown (could be done by making this bean listen for app-context shutdown for example)

documentation/en/user/source/configuration/storage.rst
nmco: Yeah this not very a nice and can be fixed with a line of code ... I will fix this.

geowebcache/sqlite/src/main/java/org/geowebcache/sqlite/FileManager.java
nmco: Nice catch !

geowebcache/sqlite/src/main/java/org/geowebcache/sqlite/MbtilesBlobStore.java
```diff
@@ -0,0 +1,538 @@
...
+    // if false we will not care about tile create time (expiration rules will not work)
+    private final boolean useCreateTime;
+
+    private final BlobStoreListenerList listeners;
+
+    // parsed mbtiles metadata indexed per layer
+    private final Map<String, MBTilesMetadata> layersMetadata = new ConcurrentHashMap<>();
+
+    // Executor that can be used to perform parallel operations
+    private final ExecutorService executorService;
```
aaime: The executor service should be shut down on application undeploy (again, you can listen to spring events to do that).

geowebcache/sqlite/src/main/java/org/geotools/mbtiles/GeoToolsMbtilesUtils.java
```diff
@@ -0,0 +1,89 @@
...
+
+        public MbtilesFileExtended(Connection connection) throws Exception {
+            super(new ExtendedDataSource(connection));
+        }
+    }
+
+    /**
+     * Extended version of a data source that always return the passed connection.
+     */
+    private final static class ExtendedDataSource extends BasicDataSource {
```
aaime: This data source is not "extended", it's actually single connection.
E.g. https://github.com/geotools/geotools/blob/master/modules/library/referencing/src/main/java/org/geotools/referencing/factory/epsg/SingleConnectionDataSource.java
(too bad it's not public)

aaime: Actually, if I read correctly you have spring-jdbc in the classpath, so you can use http://docs.spring.io/spring/docs/current/javadoc-api/org/springframework/jdbc/datasource/SingleConnectionDataSource.html

geowebcache/sqlite/src/main/java/org/geowebcache/sqlite/MbtilesBlobStore.java
```diff
@@ -0,0 +1,538 @@
...
+            }
+            Utils.debug(LOGGER, "Tile not found on file '%s'.", file);
+            return false;
+        });
+        if (exists && useCreateTime) {
+            // the tile exists and we need to set the create time in the tile object
+            Long createdTime = getTileCreateTime(file, tile.getXYZ()[2], tile.getXYZ()[0], tile.getXYZ()[1]);
+            if (createdTime == null) {
+                // no create time associated with this tile lets assume the current as the create time
+                createdTime = System.currentTimeMillis();
```
aaime: Thinking out loud, wouldn't it make more sense to use the file creation time, or the last modified file time?

geowebcache/sqlite/src/main/java/org/geowebcache/sqlite/MbtilesBlobStore.java
```diff
@@ -0,0 +1,538 @@
...
+        }
+        return exists;
+    }
+
+    @Override
+    public boolean delete(TileObject tile) throws StorageException {
+        Utils.updateLogId("DELETE-" + tile.toString());
+        File file = fileManager.getFile(tile);
+        Utils.debug(LOGGER, "Tile mapped to file '%s'.", file);
+        initDatabaseFileIfNeeded(file, tile.getLayerName(), tile.getBlobFormat());
```
aaime: Hmm... if the tile is not there do we need to init it to remove a tile that won't be inside it? :-)

geowebcache/sqlite/src/main/java/org/geowebcache/sqlite/SqliteConnectionManager.java
```diff
@@ -0,0 +1,425 @@
...
+        try {
+            Class.forName("org.sqlite.JDBC");
+        } catch (Exception exception) {
+            throw Utils.exception(exception, "Error initiating sqlite driver.");
+        }
+        // computing some values used by the pool reaper
+        double poolSizeThreshold = poolSize * 0.9;
+        double connectionsToRemove = poolSize * 0.1;
+        // starting the connection pool reaper
+        new Thread(() -> {
```
aaime: This thread should be closed at application shutdown

aaime: Also, since the eviction happens at a certain interval, it means the number of files open concurrently go above threshold for a certain amount of time. This should be documented.
I am left wondering why a guava cache was not used instead, it would have done the job of keeping the size at max all times without needing a separate thread: https://github.com/google/guava/wiki/CachesExplained
Not saying you have to rewrite it, if it works it works...

geowebcache/sqlite/src/main/java/org/geowebcache/sqlite/MbtilesBlobStore.java
nmco: Yes you are correct, I will go for the creation time ... if the we use the creation time and the file was modified it means that those tiles already have a creation time.

nmco: Changed my mind I will go for the last modified time, this will maintain a consistent behavior between windows and linux.

geowebcache/sqlite/src/main/java/org/geowebcache/sqlite/MbtilesBlobStore.java
nmco: Euhhh ... probably not :P

documentation/en/user/source/configuration/storage.rst
nmco: I see your point and agree with you. Although, {layer}{grid}{format}{params} doesn't give enough granularity I will instead go for something like this: {layer}{grid}{format}{params}/{z}-{x}-{y}.sqlite with tiles ranges 250x250 which give us 62500 tiles per file. Does this sounds good ?

documentation/en/user/source/configuration/storage.rst
nmco: Fixed.

documentation/en/user/source/configuration/storage.rst
nmco: Fixed.

documentation/en/user/source/configuration/storage.rst
nmco: Fixed.

documentation/en/user/source/configuration/storage.rst
nmco: Fixed.

documentation/en/user/source/configuration/storage.rst
nmco: Fixed.

documentation/en/user/source/configuration/storage.rst
nmco: Fixed.

documentation/en/user/source/configuration/storage.rst
nmco: Fixed.

documentation/en/user/source/configuration/storage.rst
nmco: Fixed (documentation updated).

geowebcache/sqlite/src/main/java/org/geowebcache/sqlite/Utils.java
nmco: Fixed.

geowebcache/sqlite/src/main/java/org/geowebcache/sqlite/FileManager.java
nmco: Fixed.

geowebcache/sqlite/src/main/java/org/geowebcache/sqlite/MbtilesBlobStore.java
nmco: Fixed.

geowebcache/sqlite/src/main/java/org/geowebcache/sqlite/MbtilesBlobStore.java
nmco: Fixed.

geowebcache/sqlite/src/main/java/org/geotools/mbtiles/GeoToolsMbtilesUtils.java
nmco: Correct ... switched to spring single connection.

nmco: Fixed.

geowebcache/sqlite/src/main/java/org/geowebcache/sqlite/MbtilesBlobStore.java
nmco: Fixed.

geowebcache/sqlite/src/main/java/org/geowebcache/sqlite/SqliteConnectionManager.java
nmco: I started by using Guava but at some point I realized that I wanted also to remove the less used connections and I could not find a way for doing this with guava , i.e. defining a max size and when the max size is hit remove the less used connections.

nmco: The reaper thread is close at the application shutdown.

geowebcache/web/src/main/webapp/WEB-INF/geowebcache-sqlite-context.xml
nmco: Asked for feedback in GeoServer and GeoWebCache mailing lists: http://osgeo-org.1560.x6.nabble.com/Using-spring-annotations-to-configure-REST-controllers-td5274528.html

nmco:
I updated the pull request with your suggestions ... thanks for the deep and detailed review !

geowebcache/sqlite/src/main/java/org/geowebcache/sqlite/FileManager.java
aaime: Doh, I was wrong though, thread locals should be cleaned at each request end, otherwise they will be left attached to the threads and prevent clean undeploy.

documentation/en/user/source/configuration/storage.rst
```diff
@@ -272,6 +271,175 @@ GeoServer ``topp:states`` sample layer on a fictitious ``my-geowebcache-bucket``
...
+`````````````````
+
+The MBTiles specification don't give information about when a tile was created. To allow expire rules, an auxiliary table is used to store tile creation time. In the presence of an MBTiles file generated by a third party tool it is assumed that the creation time of a tile was the first time it was accessed. This feature can be activated or deactivated by configuration. Note that this will not break the MBTiles specification compliance.
+
+Eager Truncate
+```````````````
+
+When performing a truncate of the cache the store will try to remove the whole database file avoiding to create fragmented space. This is not suitable for all the situations and is highly dependent on the database files granularity. The configuration property ``eagerDelete`` allows the user to disable or deactivate this feature which is disabled by default. 
+
+When a truncate request by tile range is received all the the databases files that contains tiles that belong to the tile range are identified. If eager delete is set to true those databases files are deleted otherwise we perform a single delete query for each file.
```
aaime: Ah hem, leftover "we" usage

geowebcache/sqlite/src/main/java/org/geowebcache/sqlite/FileManager.java
nmco: Removed the thread local.

aaime:
Thread local gone, build is green... merging

PR #390, opened 2016-06-16
Resolves #389, allow a different layer name to be advertised

### LINKED ISSUE #389: Allow the use of a custom name when advertising a tile layer
When advertising a tile layer (for example a get capabilities result) we should be able to use a custom name. By default, the layer name should be used.

The current use case for this is the support of per-workspace services in GeoServer integration.

nmco:
Associated issue: https://github.com/GeoWebCache/geowebcache/issues/389

This is needed to handle per-workspace services in GeoServer integration with GWC.

### DIFF
--- a/geowebcache/core/src/main/java/org/geowebcache/layer/TileLayer.java
+++ b/geowebcache/core/src/main/java/org/geowebcache/layer/TileLayer.java
@@ -113,6 +113,14 @@ public abstract class TileLayer {
     public abstract String getName();
 
     /**
+     * Name that should be used when advertising the layer (for example in a get capabilities result document).
+     * @return
+     */
+    public String getAdvertisedName() {
+        return getName();
+    }
+
+    /**
      * @return {@code true} if the layer is enabled, {@code false} otherwise
      */
     public abstract boolean isEnabled();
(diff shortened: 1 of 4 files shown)


aaime:
It should be recognized this is needed to handle per-workspace services, however Nuno, I believe that for this one to be acceptable in GWC land the advertised name should also be checked in the native GWC services, I might be wrong, but as part of the TileLayer fields doesn't it also become editable via xml config files?

nmco:
Thanks for warning me about the missing context of this pull request, I updated the issue. 

Regarding the the advertised attribute name. TileLayer class is abstract and doesn't have any fields and each sub-class is responsible to register the fields that are editable by configuration. 

This field would be editable by default only if I added it in AbstractTileLayer class (https://github.com/GeoWebCache/geowebcache/blob/5d841c426741ce71eb69a9b11c9e8d14ebdfc7bb/geowebcache/core/src/main/java/org/geowebcache/layer/AbstractTileLayer.java).

So the advertised field is not editable via xml config files and right now it will preserve the current behavior, i.e. use the layer name as the advertised name.

aaime:
Looking at the patch, what happens if, in a stand-alone GWC, an implementation sets an advertised name, the caps report that name, a WMTS GetTile comes in against it... how is the code finding the right tiles? I don't see the code doing that (I assume there is some code in GeoServer for the integration, but this change should also work in the stand alone version).

nmco:
With the help of Andrea Aime a better solution for this was found. That solutions was implemented in GeoServer pull request https://github.com/geoserver/geoserver/pull/1645.

Issue #392, opened 2016-06-19
Allow a tile layer to provide styles legend information


nmco:
A tile layer should be able to provided legend information that can be used to generated legend URL elements in a WMTS get capabilities result.

The ability to add extra metadata to a WMTS GetCapabilities result is need in order to achieve INSPIRE compliance.

Issue #391, opened 2016-06-19
Add extensions points in WMTS service


nmco:
In this case we only need to extend the GetCapbilities operation but the created interface (WMTSExtension) could be used to extends the other operations.

The ability to add extra metadata to a WMTS GetCapabilities result is need in order to achieve INSPIRE compliance.

PR #393, opened 2016-06-19
Add WMTS extensions points and extra metadata to WMTS GetCapabilities result

nmco:
This changes are made in the context of making WMTS compliant with INSPIRE.

Associated issues:
- https://github.com/GeoWebCache/geowebcache/issues/391
- https://github.com/GeoWebCache/geowebcache/issues/392

### DIFF
--- a/geowebcache/wmts/src/main/java/org/geowebcache/service/wmts/WMTSGetCapabilities.java
+++ b/geowebcache/wmts/src/main/java/org/geowebcache/service/wmts/WMTSGetCapabilities.java
@@ -25,6 +25,7 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.Iterator;
 import java.util.List;
+import java.util.Map;
 
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
@@ -269,6 +270,7 @@ public class WMTSGetCapabilities {
     }
 
     private void serviceIdentification(XMLBuilder xml, ServiceInformation servInfo) throws IOException {
+        
         xml.indentElement("ows:ServiceIdentification");
         
         if (servInfo != null) {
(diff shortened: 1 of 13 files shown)


geowebcache/wmts/src/main/java/org/geowebcache/service/wmts/WMTSService.java
```diff
@@ -65,6 +73,9 @@
     
     private GeoWebCacheDispatcher controller = null;
 
+    @Autowired(required = false)
```
aaime: Normally GeoWebCache loads extension points using GeoWebCacheExtensions instead of autowiring. In GWC the mechanism is not all that different from autowiring for the moment, but eventually it could grow similar to GeoServer where there is the ability to declare extensions filters and control load order via priority.

geowebcache/wmts/src/main/java/org/geowebcache/service/wmts/WMTSGetCapabilities.java
```diff
@@ -128,8 +157,119 @@ private String generateGetCapabilities(Charset encoding) {
...
+                mergeServiceInformation(servInfo, serviceInformation);
+            }
+        }
+        return servInfo;
+    }
+
+    /**
+     * Substitute serviceA information with no NULL information from serviceB.
+     */
+    private void mergeServiceInformation(ServiceInformation serviceA, ServiceInformation serviceB) {
```
aaime: This makes the code order dependent, wondering if we should add a GeoServer ExtensionPriority like interface to GWC as well?

aaime:
Added a couple of comments, looks otherwise good

nmco:
Aye aye ... I don't know how I missed the GeoWebCacheExtensions class. I add an extension priority interface (copy paste from GeoServer) and make GeoWebCacheExtensions aware of it. Thansk for the review.

aaime:
Looking good to me. It's been around two weeks without any other reviewer hitting it, merging.

PR #395, opened 2016-06-24
Support for image/vnd.jpeg-png format for tiles storage and response

aaime:
This work adds one dependency to jai-ext in core, to avoid having to put a verbatim copy of a class in GWC and GeoServer. Unsure about documentation, I could not find a section dedicated to formats.
While I have not tried, it should be possible to use jpeg-png also against a remote server that only supports PNG, the current code is splitting the meta-tile and determining on a tile per tile basis if it should be encoded as PNG or JPEG.

### DIFF
--- a/geowebcache/core/src/main/java/org/geowebcache/layer/MetaTile.java
+++ b/geowebcache/core/src/main/java/org/geowebcache/layer/MetaTile.java
@@ -20,12 +20,12 @@ import java.awt.Rectangle;
 import java.awt.RenderingHints;
 import java.awt.image.BufferedImage;
 import java.awt.image.RenderedImage;
+import java.awt.image.renderable.ParameterBlock;
 import java.io.IOException;
 import java.io.OutputStream;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashSet;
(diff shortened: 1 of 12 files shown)


geowebcache/core/src/main/java/org/geowebcache/layer/MetaTile.java
```diff
@@ -388,6 +393,32 @@ public boolean writeTileToStream(final int tileIdx, Resource target) throws IOEx
...
+            for (int i = 0; i < bands.length; i++) {
+                bands[i] = i;
+            }
+            // ParameterBlock creation
+            ParameterBlock pb = new ParameterBlock();
+            pb.setSource(ri, 0);
+            pb.set(bands, 0);
+            final RenderingHints hints = new RenderingHints(JAI.KEY_IMAGE_LAYOUT, new ImageLayout(ri));
+            ri = JAI.create("BandSelect", pb, hints);
+//            ri = new ImageWorker(ri).retainBands(3).getRenderedImage();
```
dromagnoli: I would remove this line :)

dromagnoli:
Looks good to me.

PR #402, opened 2016-07-19
Add mbtiles blobstore (backport to 1.9.x)

nmco:


aaime:
So, Kevin is good with this backport, I've checked it too, discussed with Nuno and we tried to figure out "what could possibly go wrong" but as long as people are not using the new mbtiles support, nothing should possibly break (so, upgrade safe). So +1 on the backport too.
@smithkm do you know if/when you'll be done with the security fixes? If there is a bunch of days left I'd merge right away, otherwise I guess I would wait right after the release (e.g., I don't want to merge it the day before the release, but I see no reason to make it wait a week).

nmco:
Realized that GeoServer GWC is not dependent of GWC Spring configuration. To avoid any unexpected side effects the MBTiles blobstore configuration inclusion could be commented. If someone wants to use the MBTiles blobstore he just has to uncomment that line.  Pull request updated.

Note: GeoServer plugin will contribute the correct configuration files.

geowebcache/web/src/main/webapp/WEB-INF/geowebcache-servlet.xml
```diff
@@ -15,6 +15,9 @@
   <import resource="geowebcache-wmsservice-context.xml"/>
   <import resource="geowebcache-wmtsservice-context.xml"/>
   <import resource="geowebcache-s3storage-context.xml"/>
+  <!-- Uncomment to activate SQLite based blob stores
```
nmco: MBTiles blob store beans are not included by default.

aaime:
Ah, good thinking, this makes the pull request completely safe indeed

PR #421, opened 2016-08-27
Updated Demo Request Page

tbarsballe:
- (@smithkm) Update demo list to support vector tile formats  
- (@smithkm) Replace string concatenation in demo page generator with StringBuffer…
- (@tbarsballe) Upgrade demo page generator to use OL3. Update ol.js from 3.0.0 to 3.18.0


tbarsballe:
@smithkm - I've done some basic manual testing. If you want to try out some of the more advanced parameter filters (custom ones, etc.) and other format support (vector tiles), that would be appreciated.

smithkm:
Fixes #419 #420

smithkm:
@tbarsballe Thanks, I'll look it over.

smithkm:
OK, it looks like the page itself doesn't support vector tiles, but we should be closer. Hopefully it just requires swapping out the layer class being used in OL or something like that.

smithkm:
Also, I figured out how to control the `X-Frame-Option` header that you had to work around. (#422)

tbarsballe:
For vector tiles, looks live it is `ol.layer.VectorTile` and `ol.source.VectorTile` respectively. I can add this functionality when I get back to Victoria if you want. Some of the parameters may be different as well.

smithkm:
Also there seems to be a reference to 'EPSG:4326' creeping in regardless of the gridset being used.

geowebcache/core/src/main/java/org/geowebcache/demo/Demo.java
```diff
@@ -166,195 +175,316 @@ private static String tableRows(TileLayerDispatcher tileLayerDispatcher,
...
+                + "        scale = Math.round(scale / 1000000) + 'M';\n"
+                + "    } else {\n"
+                + "        scale = Math.round(scale);\n"
+                + "    } \n"
+                + "    this.element.innerHTML = 'Scale = 1 : ' + scale;\n"
+                + "  }, this);\n"
+                + "  ol.control.Control.prototype.setMap.call(this, map);\n"
+                + "}\n"
+                + "\n"
+                + "var projectionName = 'EPSG:4326';\n"
```
tbarsballe: I think I found the erroneous EPSG:4326

smithkm:
I have the gridset issue fixed: https://github.com/smithkm/geowebcache/commit/052e5a17b3af3f7f5b02e2129ee2eac8ba59f3b1

smithkm:
Hmm, also seems to break on tile sizes other than 256x256

tbarsballe:
Added VectorTile support.

Also, the problem with different grid sizes _might_ be fixed, I was unable to test it.

tbarsballe:
@smithkm - I think this is about ready to merge now, is there anything else you think is missing?

tbarsballe:
Updated OL3 support, adding custom projection + units.

tbarsballe:
One more fix with some bounds/origins improvements.
Tested with all the default GeoWebCache layers, ought to work now (I hope).

smithkm:
This doesn't work right when embeded in GeoServer.  It can't find `ol.js` or `ol.css`

tbarsballe:
@smithkm - I tracked down the problem you had yesterday with tile formats.

Turns out it was a an issue generating the OL3 preview page wrong - when choosing whether or not to use vector tiles, it was looking at the list of supported Mime Types and grabbing the first one, rather than using the format parameter. This has now been fixed to grab the correct format parameter to the demo page.

This means that depending on what order you defined the formats in geowebcache.xml, Vector Tiles might or might not have worked.

Do you want to verify vector tiles work on your machine now?

Also, OL3 path has been fixed to work with embedded geowebcache.

PR #426, opened 2016-09-05
Resolves #425, Provides default implementations for some extension points

### LINKED ISSUE #425: WMTSExtension interface backward compatibility should be guaranteed
New introduced methods in WMTSExtension interface break backward compatibility.
Default implementations for the new extension points should be provided.

nmco:
Makes WMTSExtension interface backward compatible by providing default implementations for some extension points.


bencaradocdavies:
I can confirm that this change fixes the GeoServer build for me locally.

aaime:
Yep, also pretty important since that interface has been backported to the stable series, so just adding new methods would break it.

PR #433, opened 2016-10-03
Use cannonical path for FileManagerTest where applicable

tbarsballe:
GeoWebCache fails FileManagerTest.testPathTemplateWithTileObject() on OS X, due to confusion between `/private/var/...` and `/var/...`
This fixes that test failure.


jodygarnett:
Thank you @tbarsballe builds for me on mac now

PR #436, opened 2016-10-05
ensure caused by information included in service exception

jodygarnett:


geowebcache/wms/src/main/java/org/geowebcache/config/GetCapabilitiesConfiguration.java
```diff
@@ -137,11 +137,13 @@ public String getIdentifier() {
...
-        if (wms == null) {
-            throw new ConfigurationException("Unable to connect to " + this.url);
-        }
-
+        WebMapServer wms = null;
+        
+        try {
+            wms = getWMS();
+        } catch (ServiceException | IOException e) {
+            throw new ConfigurationException("Unable to connect to " + this.url + " :"+e.getMessage(), e);
```
DBlasby: Could you change the message here ("Unable to connect..."?  It's confusing as the actual error is likely a parse error instead of a connection issue.

jodygarnett:
I did check, the method was only called from the one location.

geowebcache/wms/src/main/java/org/geowebcache/config/GetCapabilitiesConfiguration.java
smithkm: Something like `Unable to get layers` or `Unable to read capabilities` would work.

PR #439, opened 2016-10-14
Fix GeoTools dependency (should be 17-SNAPSHOT)

bencaradocdavies:
@smithkm @aaime someone please merge this as the broken GeoTools dependency is breaking the GeoWebCache build (and will break the GeoServer build as soon as @jodygarnett branches and updates versions on master).


aaime:
thanks!

Issue #387, opened 2016-06-04
The reader selected when handling the MetaTile is random 


nmco:
In the MetaTile class the selected reader is random:
https://github.com/GeoWebCache/geowebcache/blob/5d841c426741ce71eb69a9b11c9e8d14ebdfc7bb/geowebcache/core/src/main/java/org/geowebcache/layer/MetaTile.java#L360-L361

With JAI present we may not use the JAI reader because the normal one was loaded first.

aaime:
Wondering, picking the native one in PNG might actually slow things down compared to PNGJ. I though that ImageIO was using priorities to pick the "best" writer for the job (based on declared priorities). Nuno,  did you get by any chance the PNGJ writer instead of the PNG native one?

PR #440, opened 2016-10-18
Always select native ImageWriter if available

basduineveld:
This fixes issue #387: The reader selected when handling the MetaTile is random (https://github.com/GeoWebCache/geowebcache/issues/387)


tbarsballe:
Looks like this has a merge conflict, and needs to be rebased.
The fix itself looks reasonably straightforward, although there is some discussion on #387 about using priority to select the Writer.

Also, could you please add a test case?

basduineveld:
Actually, this "straightforward fix" might just be too simplistic. The conflicting edit shows that the native writer is not suitable for writing 2-4 bit PNG files and the comment on #387 suggests that it's suboptimal in other cases as well. On the other hand, my team experienced an issue where the built-in (non-native) JPEG writer generated corrupt images when using meta tiling. So really, which writer is better is different for every situation and I think a robust fix for #387 needs more research.

tbarsballe:
Ok, thanks for the input. In that case, I'll close this PR pending a more robust fix.

PR #457, opened 2016-12-22
[fixes #427] - NIOLockProvider will not retry when failing to create output stream

### LINKED ISSUE #427: NIOLockProvider will not retry when failing to create output stream
If NIOLockProvider fails to create output stream client gets empty tiles:

HTTP error code 400: "Failure while trying to get lock for key meta_{tile}.png"

GeoWebCache v1.9.1

```
currFos = new FileOutputStream(file);

try {
  currLock = currFos.getChannel().lock();
 } catch (OverlappingFileLockException e) {
    IOUtils.closeQuietly(currFos);
...
```

Moving `currFos = new FileOutputStream(file);` into the try would solve the issue.

emilzse:
* Moved initialization of FileOutputStream into try to retry on exceptions


emilzse:
This pull is replaced with #459

smithkm:
I think you meant to close the PR rather than the issue.

smithkm:
Wait sorry this was right to begin with.

PR #459, opened 2017-01-02
[fixes #427] - NIOLockProvider will not retry when failing to create output stream

### LINKED ISSUE #427: NIOLockProvider will not retry when failing to create output stream

emilzse:
* Moved initialization of FileOutputStream into try to retry on exceptions
* Added possibility to configure waitBeforeRetry and maxLockAttempts as constructor args
* Will use waitBeforeRetry in thread sleep

### DIFF
--- a/geowebcache/core/src/main/java/org/geowebcache/locks/NIOLockProvider.java
+++ b/geowebcache/core/src/main/java/org/geowebcache/locks/NIOLockProvider.java
@@ -37,24 +37,35 @@ public class NIOLockProvider implements LockProvider {
     
     public static Log LOGGER = LogFactory.getLog(NIOLockProvider.class);
 
-    private String root;
+    private final String root;
+
     /**
      * The wait to occur in case the lock cannot be acquired
      */
-    int waitBeforeRetry = 20; 
+    private final int waitBeforeRetry;
+
     /**
      * max lock attempts
      */
-    int maxLockAttempts = 120 * 1000 / waitBeforeRetry;
+    private final int maxLockAttempts;
 
     MemoryLockProvider memoryProvider = new MemoryLockProvider();
 
     public NIOLockProvider(DefaultStorageFinder storageFinder) throws ConfigurationException {
+        this(storageFinder.getDefaultPath());
+    }
+
+    public NIOLockProvider(DefaultStorageFinder storageFinder, int waitBeforeRetry,
+            int maxLockAttempts) throws ConfigurationException {
         this.root = storageFinder.getDefaultPath();
+        this.waitBeforeRetry = waitBeforeRetry;
+        this.maxLockAttempts = maxLockAttempts;
     }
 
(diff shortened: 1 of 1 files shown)


geowebcache/core/src/main/java/org/geowebcache/locks/NIOLockProvider.java
```diff
@@ -73,21 +84,22 @@ public NIOLockProvider(String root) throws ConfigurationException {
                 while(currLock == null && count < maxLockAttempts) {
                     // the file output stream can also fail to be acquired due to the
                     // other nodes deleting the file
-                    currFos = new FileOutputStream(file);
                     try {
+                        currFos = new FileOutputStream(file);
+
                         currLock = currFos.getChannel().lock();
                     } catch(OverlappingFileLockException e) {
                         IOUtils.closeQuietly(currFos);
```
smithkm: This shouldn't be needed now that try-with-resources is being used.

geowebcache/core/src/main/java/org/geowebcache/locks/NIOLockProvider.java
```diff
@@ -73,21 +84,22 @@ public NIOLockProvider(String root) throws ConfigurationException {
...
                         try {
-                            Thread.sleep(20);
+                            Thread.sleep(waitBeforeRetry);
                         } catch (InterruptedException ie) {
                             // ok, moving on
                         }
                     } catch(IOException e) {
                         // this one is also thrown with a message "avoided fs deadlock"
                         IOUtils.closeQuietly(currFos);
```
smithkm: Likewise shouldn't be needed now

geowebcache/core/src/main/java/org/geowebcache/locks/NIOLockProvider.java
emilzse: try-with-resource is not suitable solution here since the stream should not be closed if successfully created inside the try

smithkm: OK.  Only criticism remaining then is the need for a unit test.

smithkm:
Since this is a bug fix, not a feature, it is not affected by the feature freeze.  Getting it in before the freeze would still be nice, but it can be merged after.

smithkm:
OK looked at making a unit test and it's probably not worth the trouble for this fix.

PR #461, opened 2017-01-03
Disk quota - JDBC - schema

emilzse:
* adds possibility to configure schema for jdbc quota store in geowebcache-diskquota-jdbc.xml

### DIFF
--- a/geowebcache/diskquota/jdbc/src/main/resources/org/geowebcache/diskquota/jdbc/geowebcache-diskquota-jdbc.xsd
+++ b/geowebcache/diskquota/jdbc/src/main/resources/org/geowebcache/diskquota/jdbc/geowebcache-diskquota-jdbc.xsd
@@ -21,6 +21,12 @@
             <xs:documentation>JNDI resource used to locate the connection pool, optional, if this is missing then the connection pool element must be configured</xs:documentation>
           </xs:annotation>
         </xs:element>
+        <!-- Support for schema in configuration -->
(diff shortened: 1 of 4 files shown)


emilzse:
```
<?xml version="1.0" encoding="utf-8" ?>
<gwcJdbcConfiguration>
   <dialect>PostgreSQL</dialect> 
   <JNDISource>jdbc/DBDS</JNDISource> 
   <schema>gwc</schema> 
</gwcJdbcConfiguration>
```

geowebcache/diskquota/jdbc/src/main/resources/org/geowebcache/diskquota/jdbc/geowebcache-diskquota-jdbc.xsd
```diff
@@ -21,6 +21,12 @@
             <xs:documentation>JNDI resource used to locate the connection pool, optional, if this is missing then the connection pool element must be configured</xs:documentation>
           </xs:annotation>
         </xs:element>
+        <!-- Support for schema in configuration -->
+        <xs:element name="schema" type="xs:string" minOccurs="0">
+          <xs:annotation>
+            <xs:documentation>Database schema where the tables will be created. Defaults to public. Schema must exists.</xs:documentation>
```
smithkm: The value of the configuration parameter in GWC is not actually defaulting to `public` the DBMS is defaulting to whatever it defaults to, which in the case of Postgres is `public`.  Other DBMSes default other ways or lack this idea of a 'schema'.

PR #463, opened 2017-01-03
GetCapabilities - WMS/WMTS

emilzse:
* When there are significant amount of layers GetCapabilities for WMS/WMTS can be very slow and response very big. Therefore added support to only return specific layers in GetCapabilities by providing layers name in LAYERS parameter
* Also if there are a lot of different gridsets for WMTS then specifying gridset could optimize the response. Therefore added support to only return specific gridset by providing gridset id in SRS parameter 

Some example URLs:
http://host/geowebcache/service/wms?REQUEST=getcapabilities&LAYERS=layerName
http://host/geowebcache/service/wmts?REQUEST=getcapabilities&LAYERS=layerName
http://host/geowebcache/service/wmts?REQUEST=getcapabilities&LAYERS=layerName&SRS=EPSG:4326


smithkm:
Adding a proprietary extension to an OGC protocol is really something that should be discussed in detail on the mailing list.

emilzse:
2017-01-04 kl. 13:44, Emil Zail:
>
> Hi,
>
> When GWC is configured with many layers the GetCapabilities response for WMS/WMTS is both slow and very big.
>
> Below are some numbers for GetCapabilities requests to GWC where we have configured around 430 layers. The numbers are of course very rough but gives you a hint. (Download time is excluded from waiting time).
>
> WMS
> Waiting time: > 750ms
> Size: 1.6 MB
>
> WMTS
> Waiting time: > 5s
> Size: 115 MB
>
> In many cases there is just a need to get response with one layer and in case of WMTS only one gridset.
>
> Therefore we would like to extend the GetCapabilities endpoint with optional parameters LAYERS and SRS (WMTS).
>
> LAYERS: A comma separated list of layer names
> SRS (WMTS): Gridset ID
>
> Below are numbers for response with one layer specified with the LAYERS parameter
>
> WMS
> Waiting time: > 5ms
> Size: 5 KB
>
> WMTS
> Waiting time: > 100ms
> Size: 2.3 MB
> SRS provided
> Waiting time: > 5ms
> Size: 32 KB
>
> How the URLs could look like:
> http://host/geowebcache/service/wms?REQUEST=getcapabilities&LAYERS=layerName
> http://host/geowebcache/service/wmts?REQUEST=getcapabilities&LAYERS=layerName
> http://host/geowebcache/service/wmts?REQUEST=getcapabilities&LAYERS=layerName&SRS=EPSG:4326
>
> The parameters are optional and it will always fallback to responding with all.
>
> Best regards
>
> Emil

smithkm:
OK, Given that the OGC is working on ideas for doing this according to @aaime's email, I'm inclined to OK this as a stopgap solution until they make their minds up.  However I would like to be clear about that and avoid any potential future conflict with whatever they come up with so I'd like to change the name of the parameter to something else that is safer.  `VND_LAYERS`, `X_LAYERS`, or `GWC_LAYERS` are possibilities.  That will make it more clear that this is an ad hoc thing we've added and the OGC is unlikely to add anything by those names so w should be safe for future updates.

I'd also like some unit test coverage for the new behaviour.  Some documentation in the user guide explaining the new feature should also be included, but docs be done during feature freeze so I can let that slide.

emilzse:
+1 for GWC_LAYERS

aaime:
@emilzse any chance you can solve the conflicts?
OGC in the meantime did not go for a caps filtering because new OGC API came up, the issue is really still there, in the collections resource, but eventually an extension of it might help address it.

aaime:
PR stuck, closing.

PR #465, opened 2017-01-03
fullWMS stitching format

emilzse:
* fixes issue where fullWMS request for png would generate tiles in png8
* will use png for stiching-tiles to jpeg
* will use requested format for stiching-tiles if supported

### DIFF
--- a/geowebcache/wms/src/main/java/org/geowebcache/service/wms/WMSTileFuser.java
+++ b/geowebcache/wms/src/main/java/org/geowebcache/service/wms/WMSTileFuser.java
@@ -272,19 +272,26 @@ public class WMSTileFuser{
         
         ImageMime firstMt = null;
         
-        if(iter.hasNext()){
-            firstMt = (ImageMime)iter.next();
-        }
-        boolean outputGif = outputFormat.getInternalName().equalsIgnoreCase("gif");
+        boolean outputJpeg = outputFormat.getInternalName().equalsIgnoreCase("jpeg");
         while (iter.hasNext()) {
-            MimeType mt = iter.next();           
-            if(outputGif){
-                if (mt.getInternalName().equalsIgnoreCase("gif")) {
+            MimeType mt = iter.next();
+
+            // Sets first here so that it will be part of loop
+            if (firstMt == null) {
+                firstMt = (ImageMime) mt;
+            }
+
+            if (outputJpeg) {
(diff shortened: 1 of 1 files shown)


geowebcache/wms/src/main/java/org/geowebcache/service/wms/WMSTileFuser.java
```diff
@@ -272,19 +272,26 @@ protected WMSTileFuser(TileLayerDispatcher tld, StorageBroker sb, HttpServletReq
...
+            MimeType mt = iter.next();
+
+            // Sets first here so that it will be part of loop
+            if (firstMt == null) {
+                firstMt = (ImageMime) mt;
+            }
+
+            if (outputJpeg) {
+                // use png for stitching to jpeg
+                if (mt.getFileExtension().equalsIgnoreCase("png")) {
```
smithkm: Using an arbitrary PNG subtype?  It seems like preference should be given to truecolour PNG over indexed png Internal names `png` vs `png8`  Using the extension will admittedly ensure you get one of the PNG types, but it is possible, if an unusal decision on the part of the server admin, that both are available.

geowebcache/wms/src/main/java/org/geowebcache/service/wms/WMSTileFuser.java
```diff
@@ -272,19 +272,26 @@ protected WMSTileFuser(TileLayerDispatcher tld, StorageBroker sb, HttpServletReq
...
                     this.srcFormat = (ImageMime) mt;
                     break;
                 }
             }
-            if (mt.getInternalName().equalsIgnoreCase("png")) {
+
+            // if supported format matches requested format use it
+            if (mt.getMimeType().equalsIgnoreCase(outputFormat.getMimeType())
+                    && mt.getFileExtension().equalsIgnoreCase(outputFormat.getFileExtension())) {
```
smithkm: Why the extension check in addition to comparing mime types?

emilzse: I found that the object MimeType for png and png8 would return image/png from getMimeType(). Therefore had to also check the file extension( ".png" vs ".png8"). 

Is there a better way for to handle this?

aaime:
No follow up and accumulated conflicts, closing

PR #467, opened 2017-01-04
Fix #466. Fix variable mixup in MimeType.isCompatible

### LINKED ISSUE #466: GWC writes service exception XML when response code is 200
When GeoServer returns a service exception with response code 200 GWC will write out the XML as a tile without throwing an exception.

The check for Content-Type matching the requested image type falsely succeeds due to a bug in MimeType.isCompatible.

dvntucker:
Fixed a mistaken variable in isCompatible that was causing false positives. XML error MimeType was  reporting as compatible with image mime types, which caused GWC to mistakenly save service exceptions as images.

### DIFF
--- a/geowebcache/core/src/main/java/org/geowebcache/mime/MimeType.java
+++ b/geowebcache/core/src/main/java/org/geowebcache/mime/MimeType.java
@@ -195,10 +195,16 @@ public class MimeType {
         return format.hashCode();
     }
 
-
-
+    /**
+     * Determine whether otherMimeType is compatible with this MimeType. They're compatible if
+     * they're identical, or presumably if otherMimeType has this mime type as a prefix.
+     * @param otherMimeType the mime type to check for compatibility
+     * @return whether otherMimeType is "compatible" with this mime type
+     */
     public boolean isCompatible(String otherMimeType) {
-        return mimeType.equalsIgnoreCase(otherMimeType) || otherMimeType.startsWith(otherMimeType);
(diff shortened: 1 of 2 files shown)


geowebcache/core/src/main/java/org/geowebcache/mime/MimeType.java
```diff
@@ -198,7 +198,8 @@ public int hashCode() {
 
 
     public boolean isCompatible(String otherMimeType) {
```
smithkm: Could you update the javadoc to clarify how this is used.  Something like "Checks if the given MIME type is a subtype of this one."

geowebcache/core/src/main/java/org/geowebcache/mime/MimeType.java
```diff
@@ -198,7 +198,8 @@ public int hashCode() {
...
-        return mimeType.equalsIgnoreCase(otherMimeType) || otherMimeType.startsWith(otherMimeType);
+        return mimeType.equalsIgnoreCase(otherMimeType) ||
+            otherMimeType.startsWith(mimeType);
```
smithkm: It looks like this makes it case insensitive if it's otherwise an exact match, but case sensitive if matching a subtype, which is really an odd thing to do.

dvntucker:
Updated to have JavaDoc, and do a case-insensitive prefix check (also a null check before that to prevent NPE).

I also changed the test case to use JPEG, since I just noticed that PNG actually already has its own implementation of isCompatible which... maybe isn't needed any longer?

Issue #456, opened 2016-12-20
Writing 4-bit PNG with native JAI fails


dvntucker:
In certain circumstances using PNG8 in GeoServer the resulting GetMap result can actually be optimized to 4-bit paletted PNG. Unfortunately the native JAI PNG writer can't encode anything other than 1, 8 or 16bit PNGs so the tile fails. GeoServer gets around this by checking the bit depth when native JAI is enabled and falling back to the pure Java encoder. 

I think we could do something similar with GWC. In ImageMime.getWriter we could check for non-supported bit depth and the native PNG writer (somehow) and either fallback to the pure Java PNGWriter manually, or just move on to the next writer found in the service registry (presumably the next one would be the pure Java version anyway).

dvntucker:
GeoServer had a related ticket here:

https://osgeo-org.atlassian.net/browse/GEOT-4540

I don't think the linked PR was ever merged, but this case does appear to be handled in the PNGMapResponse

https://github.com/geoserver/geoserver/blob/master/src/wms/src/main/java/org/geoserver/wms/map/PNGMapResponse.java#L111

smithkm:
https://github.com/GeoWebCache/geowebcache/pull/469

PR #469, opened 2017-01-04
#456 Check for 2-4bit PNGs and skip the native writer if that's what comes up

dvntucker:
This is a bit of a corner case, but 2 and 4bit PNGs don't work with Native JAI. When using PNG8 with GeoServer sometimes the PNGs get encoded as 4bit PNGs as an optimization.

The solution is a bit hacky, but not too far off of how GeoServer/GeoTools handle this incompatibility. In GeoTools the native SPI is found by name and asked if it can support the RenderedImage being handled. That was a bit heavy weight for GWC, so in this case I just check for 2/4 bit PNGs and if the writer returned matches the native version (just going by class name, since we can't statically link against that class for an instanceof check) it will skip it and move on to the next writer in the list.

GS/GT's handling of the same edge case can be found in PNGMapResponse/ImageWorker respectively.


geowebcache/core/src/test/java/org/geowebcache/layer/wms/MetaTileTest.java
```diff
@@ -46,7 +46,7 @@ public void test1MetaTile() throws Exception {
                 30);
         
         long[] gridPos = { 0, 0, 0 };
-        
+
         //int[] gridBounds, int[] tileGridPosition, int metaX, int metaY
```
smithkm: Unnecessary whitespace change, but minor so we can live with it.

geowebcache/core/src/test/java/org/geowebcache/mime/ImageMimeTest.java
```diff
@@ -114,4 +120,17 @@ public void testIsCompatileFailsOnXMLMime() {
...
             pngMimeType.isCompatible("application/vnd.ogc.se_xml"));
     }
+
+    @Test
+    public void test4BitPNG() throws IOException, URISyntaxException {
+        URL url = this.getClass().getResource("/images/4bit.png");
+        RenderedImage tile = ImageIO.read(new File(url.toURI()));
+        ImageWriter writer = ImageMime.png8.getImageWriter(tile);
+        //Not a great test, but make sure our writer is the pure java one, not the native version
+        //since the native version can't handle 4bit pngs.
```
smithkm: Agreed. This is probably the best that's practical, even if it's far from ideal.  All I can think of as an alternative would be wrapping the `javax.imageio.ImageIO.getImageWritersByFormatName(internalName);` call in a method that could be overridden for testing and then providing mocks, which still has the problem of providing a mock for a class that may or may not exist where the specific class name is the important detail.

PR #482, opened 2017-02-10
480 Update pom to jt-utilities 1.0.13

ugomos:
Since jai-ext 1.0.13 is not dependant on jaitools, a previous exclusion of a jaitools module in gwc would not be valid anymore (and xerces breaking some tests was back). Therefore, the poms were modified accordingly.


aaime:
Travis pass, let's get the geoserver build fixed

PR #484, opened 2017-02-16
Resolves #483, Fix disk quota JDBC store layer gridset deletion

### LINKED ISSUE #483: Disk quota value becomes negative
To reproduce this issue just create a cache (by seeding or requesting) for two different layers using the same gridset with disk quota enabled. Empty the cache of one of the layers, the disk quota value will drop to zero (if you only had those two layers), empty the cache of the remaining layer, the disk quota value will become a negative number.

This issue happens because the ``JDBCQuotaStore`` uses the quotas of all the layers that use the provided gridset instead of the target layer:
https://github.com/GeoWebCache/geowebcache/blob/master/geowebcache/diskquota/jdbc/src/main/java/org/geowebcache/diskquota/jdbc/JDBCQuotaStore.java#L318-L335

This problem have also been reported in GeoServer: https://osgeo-org.atlassian.net/browse/GEOS-5615.

nmco:
Associated issue: https://github.com/GeoWebCache/geowebcache/issues/483

This PR fix the negative disk quota issue and update tests code to make them test the situation that was provoking this issue.

I have manually tested this with the embedded H2 database and with a PostgresSQL database and everything works as expected.

This problem have also been reported in GeoServer: https://osgeo-org.atlassian.net/browse/GEOS-5615.


aaime:
Yep, looks good. Let's wait for Travis to complete the build.

Issue #488, opened 2017-02-24
WMTS Capabilities doesn't list default style when there are other styles 


ianturton:
In the WMTS getCapabilities response for a layer with multiple styles one of which is default the default style is not listed - so that style can't be used in (for example) QGIS.

Initially reported at [GEOS-8001](https://osgeo-org.atlassian.net/projects/GEOS/issues/GEOS-8001).

PR #489, opened 2017-02-24
Fix for issue 488, default styles missing from WMTS capabilities

ianturton:
see https://github.com/GeoWebCache/geowebcache/issues/488


smithkm:
Although I can see how the changed behaviour could be useful in some situations, it also seems like the current behaviour is more correct.  It's conceivable the GWC admin might want to have a default that's not advertised as a legal value.  If they do want it advertised they can include it, and if not they can leave it off the list.  This takes that option away.

aaime:
@ianturton shall we close this one?

PR #492, opened 2017-03-02
Fix #490 Log the result of cache-lookups on the server

### LINKED ISSUE #490: How to know/log if a wms tile request was satisfied with a cache hit, or miss?
If this logging configuration is covered by the documentation then I can't find it.

Using a standalone Geowebcache in Tomcat, the access log only shows 200 requests and not whether it was a cache hit or miss. 

I've changed  ./WEB-INF/classes/log4j.properties with 
```
log4j.rootLogger=debug, info, stdout, logfile
```
but catalina.out also doesn't help.

I need to know this information to confirm whether our WMS layer seeding strategy is working.

julian1:
I'd like advice on the the best way to do this. 

We currently use squid as a tile cache but intend to move to Geowebcache for its protocol support of WMS. 

We would like to log the result of all tile requests - in much the same that we currently do with squid. 

This log output could be controlled (eg. turned-off by default) with a boolean set in config - so that other end-users would not be exposed to the change. However, as a code convention, I don't think logging statements should be wrapped up in conditional logic.

Alternatively, I could use debug() or trace() so that the output is normally suppressed. But, that doesn't seem right either - since it would require us (or someone wanting this information) to set a debug log level as part of our production configuration.

### DIFF
--- a/geowebcache/core/src/main/java/org/geowebcache/conveyor/ConveyorTile.java
+++ b/geowebcache/core/src/main/java/org/geowebcache/conveyor/ConveyorTile.java
@@ -243,6 +243,8 @@ public class ConveyorTile extends Conveyor implements TileResponseReceiver {
                 this.setCacheResult(CacheResult.MISS);
             }
 
+            log.info(this.toString() + " " + this.getCacheResult());
+
             return ret;
 
         } catch (StorageException se) {


smithkm:
Bringing this question up on the developer mailing list would probably be a good idea.  It should attract more attention there than as a PR.

julian1:
Thanks @smithkm. I subscribed the the devlist a couple of days ago. Can I get a status update on when I might be approved to post.

Issue #479, opened 2017-02-07
MBTiles blob store does not include size in delete events


smithkm:
MBTilesBlobStore does not include the size of deleted tiles when it calls its listeners when it deletes objects from its storage.

smithkm:
Discovered while working on #170

Issue #359, opened 2015-11-30
make GWC possible to truncate tiles inside a bounding box for all parameterFilters combination 


gistuanson:
- User case: Our system has multiple element types (more than 200 element types). We created a layer control to allow user to select which element types they want to see on the map. To make the performance better, we set up a theme in GWC with a parameterFilter call "elementtypes". Each time user modifies an element, we would like GWC to be able to remove cached tiles within element's bounding box for all of "elementtypes" parameterFilter combination.
- Problems: 
  - As we can imagine, creating truncation request for each combination of "elementtypes" is impossible.
  - Also, using "masstruncation" REST Api in this case is not a good way as it will remove all the cache of layer even when user modifies just a small element.

PR #505, opened 2017-03-24
Parameter reverse mapping and truncation.  Fixes  #170, #359, #479

### LINKED ISSUE #170: Truncate unused Parameter Values
For parameter filters that return non-null legalValues, it'd be nice if GWC would compare the values returned to those actually cached and truncate any not in the legal set.

Since legalValues might be generated dynamically by the ParameterFilter object this would either require some sort of periodic polling, some mechanism for other code that knows when legal values might have changed to request a truncation, or possibly an option to initiate via the Seed/Truncate system's Web and REST interfaces.

This would make it easier to manage large and changing sets of parameter values, and new specialized parameterFilters like a filter for the TIME dimension that caches each of the last 6 weeks without needing manual updates.

smithkm:
Adds the ability to track the parameter values that produced a particular parameter ID in each of the blob stores as well as the ability to delete parameters based on both ID and values.

Adds a new parameter delete event and updates each of the disk quota stores to handle it.

Adds new Mass Truncate commands to make use of this new underlying capability to delete specific parameters, automatically purge all parameters from a layer that do not match its parameter filters (#170), and to truncate tiles within a bounding box across all combinations of parameters and image formats (#359).

Expanded unit tests for blob stores to ensure consistent behaviour which uncovered unrelated bug #479 which I have also fixed.

### DIFF
--- a/geowebcache/core/src/main/java/org/geowebcache/filter/parameters/ParametersUtils.java
+++ b/geowebcache/core/src/main/java/org/geowebcache/filter/parameters/ParametersUtils.java
@@ -1,0 +1,115 @@
+package org.geowebcache.filter.parameters;
+
+import java.io.UnsupportedEncodingException;
+import java.net.URLDecoder;
+import java.net.URLEncoder;
+import java.nio.charset.StandardCharsets;
+import java.util.Arrays;
+import java.util.Comparator;
+import java.util.Map;
+import java.util.Map.Entry;
+import java.util.SortedMap;
+import java.util.TreeMap;
+import java.util.function.Function;
+import java.util.function.Predicate;
+import java.util.stream.Collectors;
+
+import org.geowebcache.storage.blobstore.file.FilePathGenerator;
+
+public class ParametersUtils {
+    
+    /**
+     * 
+     * This is not safe to parse as it does not do escaping.  It is used for the parameters ID hash
+     * to maintain compatibility with old caches.  For any other uses, {@link getKVP} is preferred.
+     * 
+     * @param parameters
+     * @return
+     */
+    public static String getLegacyParametersKvp(Map<String, String> parameters) {
+        StringBuilder sb = new StringBuilder();
+        SortedMap<String, String> sorted = new TreeMap<String, String>(parameters);
+        for (Map.Entry<String, String> e : sorted.entrySet()) {
+            if(sb.length() == 0) {
+                sb.append("?");
+            } else {
+                sb.append("&");
+            }
+            sb.append(e.getKey()).append('=').append(e.getValue());
+        }
+        String paramtersKvp = sb.toString();
+        return paramtersKvp;
+    }
+    
+    private static String encUTF8(String s) {
+        try {
+            // This would be much nicer if URLEncoder.encode took a charset object instead of a string
+            return URLEncoder.encode(s, StandardCharsets.UTF_8.name());
+        } catch (UnsupportedEncodingException e) {
+            throw new IllegalStateException("Standard encoding UTF-8 is missing");
+        }
+    }
+    
+    private static String decUTF8(String s) {
+        try {
+            // This would be much nicer if URLDecoder.decode took a charset object instead of a string
+            return URLDecoder.decode(s, StandardCharsets.UTF_8.name());
+        } catch (UnsupportedEncodingException e) {
+            throw new IllegalStateException("Standard encoding UTF-8 is missing");
+        }
+    }
+    
+    /**
+     * Create a comparator that compares the results of a function applied to both its arguments.
+     * @param derivation A function taking a T and returning a Comparable
+     * @return A comparator that compares the result of derivation applied to both arguments.
+     */
+    static <T, U extends Comparable<U>> Comparator<T> derivedComparator(Function<T, U> derivation) {
+        return (t1,t2)->derivation.apply(t1).compareTo(derivation.apply(t2));
+    }
+    
+    /**
+     * Turns the parameter list into a sorted KVP string
+     * 
+     * @param parameters
+     * @return
+     */
+    public static String getKvp(Map<String, String> parameters) {
+        return parameters.entrySet().stream()
+            .sorted(derivedComparator(Entry::getKey))
+            .map(e->String.join("=", encUTF8(e.getKey()), encUTF8(e.getValue())))
+            .collect(Collectors.joining("&"));
+    }
+    
+    /**
+     * Turns the a sorted KVP string into a parameter map.
+     * 
+     * This should only be used for parsing strings created by {@link getSafeParametersKvp} not for 
+     * parsing raw query strings
+     * 
+     * @param parameters
+     * @return
+     */
+    public static Map<String, String> getMap(String kvp) {
+        return Arrays.stream(kvp.split("&"))
+            .filter(((Predicate<String>)String::isEmpty).negate())
+            .map(pair->pair.split("=", 2))
+            .collect(Collectors.toMap(p->decUTF8(p[0]), p->decUTF8(p[1])));
+    }
+    
+    
+    /**
+     * Returns the parameters identifier for the given parameters map
+     * @param parameters
+     * @return
+     */
+    public static String getId(Map<String, String> parameters) {
+        if(parameters == null || parameters.size() == 0) {
+            return null;
+        }
+        String parametersKvp = getLegacyParametersKvp(parameters);
+        return FilePathGenerator.buildKey(parametersKvp);
+    }
+    
+    
+}
--- a/geowebcache/core/src/main/java/org/geowebcache/filter/parameters/ParametersUtils.java
+++ b/geowebcache/core/src/main/java/org/geowebcache/filter/parameters/ParametersUtils.java
@@ -31,14 +31,15 @@ import java.util.function.Function;
 import java.util.function.Predicate;
 import java.util.stream.Collectors;
 
-import org.geowebcache.storage.blobstore.file.FilePathGenerator;
+import org.apache.commons.codec.digest.DigestUtils;
 
 public class ParametersUtils {
     
     /**
      * 
-     * This is not safe to parse as it does not do escaping.  It is used for the parameters ID hash
-     * to maintain compatibility with old caches.  For any other uses, {@link getKVP} is preferred.
+     * This should be treated as an opaque Identifier and should not be parsed, it is used to 
+     * to maintain compatibility with old caches.  For any other uses, {@link getKVP} is preferred
+     * as it uses safe escaping of values.
      * 
      * @param parameters
      * @return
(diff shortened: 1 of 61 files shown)


geowebcache/core/src/main/java/org/geowebcache/storage/BlobStore.java
```diff
@@ -136,6 +207,55 @@ public boolean deleteByGridsetId(final String layerName, final String gridSetId)
 	 */
 	public boolean layerExists(String layerName);
 
+    Map<String,Optional<Map<String, String>>> getParametersMapping(String layerName);
```
aaime: New interface method without a javadoc

geowebcache/core/src/main/java/org/geowebcache/storage/BlobStore.java
```diff
@@ -136,6 +207,55 @@ public boolean deleteByGridsetId(final String layerName, final String gridSetId)
...
+                                        if(Objects.isNull(value)) {
+                                            return true; // No parameter for this filter so purge
+                                        }
+                                        return !pfilter.isFilteredValue(value); // purge if it's not a filtered value
+                                    });
+                            })
+                        .orElse(true); // Don't have the original values so purge
+                    })
+                .map(Map.Entry::getKey)
+                .map(id->{
```
aaime: The formatting seems off in many places, mostly lack of spaces were I'd exepct to find some (compared to gt/gs conventions), not sure if GWC is following the same rules though.

geowebcache/core/src/main/java/org/geowebcache/storage/blobstore/file/FilePathGenerator.java
```diff
@@ -107,42 +106,29 @@ public File tilePath(TileObject tile, MimeType mimeType) {
...
      * @return
+     * @deprecated Use {@link ParametersUtils#getParametersId(Map<String, String>)} instead
      */
     public static String getParametersId(Map<String, String> parameters) {
-        if(parameters == null || parameters.size() == 0) {
-            return null;
-        }
-        String parametersKvp = getParametersKvp(parameters);
-        return buildKey(parametersKvp);
+        return ParametersUtils.getId(parameters);
```
aaime: This calls the utils which in turn calls back this class... seems a bit weird to me... wondering if it was not best to avoid making buildKey public and just call the utils there too (like, having utils expose a buildKey method and delegate). Maybe call DigestUtils.sha1Hex there.

geowebcache/core/src/main/java/org/geowebcache/storage/blobstore/memory/MemoryBlobStore.java
```diff
@@ -742,4 +757,51 @@ public boolean executeOperation(BlobStore store, Object... objs)
...
+    @Override
+    public boolean deleteByParametersId(String layerName, String parametersId)
+            throws StorageException {
+        componentsStateLock.lock();
+        try {
+            if (LOG.isDebugEnabled()) {
+                LOG.debug("Removing Layer: " + layerName);
+            }
+            // Remove the layer from the cacheProvider
+            cacheProvider.removeLayer(layerName);
```
aaime: Removes the entire layer? I guess it would have been too hard to remove just the params, but could you elaborate? Maybe in the code comment itself?

geowebcache/core/src/main/java/org/geowebcache/storage/BlobStore.java
```diff
@@ -136,6 +207,55 @@ public boolean deleteByGridsetId(final String layerName, final String gridSetId)
...
+     * If the given layer is cached, remove 
+     * @param layer
+     * @throws StorageException
+     */
+    public default boolean purgeOrphans(TileLayer layer) throws StorageException {
+        // TODO maybe do purging based on gridset and format
+        try{
+            final List<ParameterFilter> parameterFilters = layer.getParameterFilters();
+            
+            return getParametersMapping(layer.getName()).entrySet().stream()
```
aaime: This seems like too many nested streams and lambda, I have no idea of what it does. Could it be split in more understandable building blocks?

geowebcache/core/src/main/java/org/geowebcache/storage/BlobStoreListenerList.java
```diff
@@ -27,88 +27,88 @@ public synchronized boolean removeListener(BlobStoreListener listener) {
     }
 
     public void sendLayerDeleted(String layerName) {
-        if (listeners.size() > 0) {
-            for (int i = 0; i < listeners.size(); i++) {
-                listeners.get(i).layerDeleted(layerName);
-            }
-        }
+        listeners.stream().forEachOrdered(listener->{
```
aaime: Seems me a 

````
listeners.forEach(listener -> layerDeleted(layerName));
````

would have been at the same time more compact, readable, and faster (streams are very slow)

File: geowebcache/core/src/main/java/org/geowebcache/storage/CompositeBlobStore.java [hunk 7]
```diff
@@ -120,62 +123,32 @@ public CompositeBlobStore(TileLayerDispatcher layers,
 
     @Override
     public boolean delete(String layerName) throws StorageException {
-        configLock.readLock().lock();
-        try {
-            return store(layerName).delete(layerName);
-        } finally {
-            configLock.readLock().unlock();
-        }
+        return readFunctionUnsafe(()->store(layerName).delete(layerName));
```
aaime: Nice refactor :-)

geowebcache/core/src/main/java/org/geowebcache/storage/StorageBroker.java
```diff
@@ -59,5 +84,19 @@ public abstract boolean rename(String oldLayerName, String newLayerName)
     public abstract boolean getTransient(TileObject tile);
 
     public abstract void putTransient(TileObject tile);
+    
+    public abstract Set<String> getCachedParameterIds(String layerName) throws StorageException;
```
aaime: Javadoc for the new methods? :-)

geowebcache/diskquota/core/src/main/java/org/geowebcache/diskquota/storage/TilePageCalculator.java
```diff
@@ -99,48 +111,59 @@ public BigInteger getTilesPerPage(TileSet tileSet, int zoomLevel) {
...
+                throw new IllegalArgumentException(e);
             }
+        } else {
+            tileLayer = null;
         }
-
-        return layerTileSets;
+        
+        // Loop over the gridset names, getting them from the layer if they were not provided
+        return gridSubsets
```
aaime: Another bit that has too much density/nesting imho, splitting it a bit would make it more readable.

geowebcache/diskquota/jdbc/src/main/java/org/geowebcache/diskquota/jdbc/SQLDialect.java
```diff
@@ -185,6 +185,17 @@ public String getLayerGridDeletionStatement(String schema, String layerNameParam
 
         return sb.toString();
     }
+    public String getLayerParametersDeletionStatement(String schema, String layerNameParam,
```
aaime: One blank line before the new method

geowebcache/core/src/main/java/org/geowebcache/seed/MassTruncateRequest.java
```diff
@@ -19,5 +20,5 @@
      * @return {@literal true} if successful, {@literal false} otherwise
      * @throws StorageException
      */
-    public boolean doTruncate(StorageBroker sb, Configuration config) throws StorageException;
+    public boolean doTruncate(StorageBroker sb, TileBreeder breeder) throws StorageException, GeoWebCacheException;
```
smithkm: No, this is decidedly not intended for backporting.

geowebcache/core/src/main/java/org/geowebcache/filter/parameters/ParametersUtils.java
```diff
@@ -0,0 +1,132 @@
...
+import java.util.function.Predicate;
+import java.util.stream.Collectors;
+
+import org.geowebcache.storage.blobstore.file.FilePathGenerator;
+
+public class ParametersUtils {
+    
+    /**
+     * 
+     * This is not safe to parse as it does not do escaping.  It is used for the parameters ID hash
```
smithkm: Maybe "not safe to use as anything but an opaque identifier"?

geowebcache/core/src/main/java/org/geowebcache/storage/blobstore/file/FileBlobStore.java
```diff
@@ -723,4 +757,102 @@ private long padSize(long fileSize) {
...
+            String target = filteredLayerName(layerName) + "_" + parameterCache.getName();
+            stageDelete(parameterCache, target);
+        }
+        
+        listeners.sendParametersDeleted(layerName, parametersId);
+        
+        return true;
+    }
+    
+    class CarrierException extends RuntimeException {
```
smithkm: Ah right that was part of an earlier approach to something.

geowebcache/core/src/main/java/org/geowebcache/storage/blobstore/memory/MemoryBlobStore.java
smithkm: As I recall I tried to figure out a more targeted way to do this, but MemoryBlobStore didn't provide it and adding it would have required digging into the internals of both of the cache implementations.  This PR was far reaching enough as it is, and it seemed to me memory blob store being inherently more transient than other store types this was an acceptable compromise.

geowebcache/core/src/main/java/org/geowebcache/storage/CompositeBlobStore.java
smithkm: I do like my lambdas

geowebcache/core/src/main/java/org/geowebcache/storage/BlobStore.java
smithkm: These are chaining on after the `.filter` so they are at the same level.  I think maybe I should have pulled the `})` back a level.

smithkm:
@aaime Review much appreciated.  It's been a while so I'll try going over it myself and hopefully the time away will help me be more critical.  Between your look at it and a self review, I think it should be reasonable coverage.

geowebcache/core/src/main/java/org/geowebcache/storage/blobstore/file/FilePathGenerator.java
smithkm: Good point, fixed.

smithkm:
Found and fixed a few more instances of missing javadocs and unused imports and variables

smithkm:
@aaime Did my changes satisfy your review?

aaime:
Yes, it does, sorry I lost track of this one.

PR #511, opened 2017-05-12
Squashed commit of the following:

engaric:
A blob store to allow a layered n-tier blob storage hierarchy.
    Allow id references in geowebcache.xml

    This allows the use of nestedblobstore configuration as follows

    <NestedBlobStore>
    <id>files3</id>
    <frontStoreConfig reference="file" />
    <backingStoreConfig reference="s3" />
    <enabled>true</enabled>
    </NestedBlobStore>
    </blobStores>


engaric:
Missed the update to upstream/master.  Will rework and send a new pull request.

engaric:
New pull request raised compatible with the updated parameter functionality of blobstore.

PR #514, opened 2017-05-15
Configurable S3 access control.

engaric:
ACL defaults to public as currently but can be configured as private which changes it to BucketOwnerFullControl

### DIFF
--- a/geowebcache/georss/pom.xml
+++ b/geowebcache/georss/pom.xml
@@ -54,6 +54,22 @@
       <scope>test</scope>
       <classifier>tests</classifier>
     </dependency>
+    <dependency>
+      <groupId>org.easymock</groupId>
+      <artifactId>easymockclassextension</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.easymock</groupId>
+      <artifactId>easymock</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.mortbay.jetty</groupId>
+      <artifactId>jetty</artifactId>
+      <version>6.1.20</version>
+      <scope>test</scope>
+    </dependency>
   </dependencies>
 
   <build>
--- a/geowebcache/georss/pom.xml
+++ b/geowebcache/georss/pom.xml
@@ -54,22 +54,6 @@
       <scope>test</scope>
       <classifier>tests</classifier>
     </dependency>
-    <dependency>
-      <groupId>org.easymock</groupId>
-      <artifactId>easymockclassextension</artifactId>
-      <scope>test</scope>
-    </dependency>
-    <dependency>
-      <groupId>org.easymock</groupId>
-      <artifactId>easymock</artifactId>
-      <scope>test</scope>
-    </dependency>
-    <dependency>
-      <groupId>org.mortbay.jetty</groupId>
-      <artifactId>jetty</artifactId>
-      <version>6.1.20</version>
-      <scope>test</scope>
-    </dependency>
   </dependencies>
 
   <build>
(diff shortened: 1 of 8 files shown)


geowebcache/georss/pom.xml
```diff
@@ -54,6 +54,22 @@
...
+      <groupId>org.easymock</groupId>
+      <artifactId>easymock</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.mortbay.jetty</groupId>
+      <artifactId>jetty</artifactId>
+      <version>6.1.20</version>
+      <scope>test</scope>
+    </dependency>
```
smithkm: Why are you adding dependencies to the GeoRSS module?

geowebcache/s3storage/src/test/java/org/geowebcache/s3/S3BlobStoreConfigTest.java
```diff
@@ -0,0 +1,25 @@
+package org.geowebcache.s3;
+
+import static org.junit.Assert.*;
+
+import org.junit.Test;
+
+import com.amazonaws.services.s3.model.CannedAccessControlList;
+
+public class S3BlobStoreConfigTest {
```
smithkm: It would be good to test this with deserialization from XML.  XStream is where problems with this sort of thing are most likely to occur.

geowebcache/s3storage/src/main/java/org/geowebcache/s3/S3BlobStoreConfig.java
```diff
@@ -266,6 +269,28 @@ public String getProxyPassword() {
...
         this.proxyPassword = proxyPassword;
     }
+    
+    /**
+     * Checks access type
+     *
+     * @return public or private access
+     */
+    public CannedAccessControlList getAccessControlList() {
+        return accessControlList;
```
smithkm: You might be returning null here if it was never set.  The code that calls it never does a null check.  You should either set a default here, or mark this as potentially null using the `@Nullable` annotation or the `Optional` wrapper type and then check and handle the default situation.

geowebcache/georss/pom.xml
engaric: Sorry that was staged in error.  Unrelated work.

geowebcache/s3storage/src/main/java/org/geowebcache/s3/S3BlobStoreConfig.java
engaric: Sure, I will sort that out.

engaric:
I will add the user documentation.  I assume this if the file geowebcache/documentation/en/user/source/configuration/storage.rst ?

geowebcache/s3storage/src/main/java/org/geowebcache/s3/S3BlobStoreConfig.java
```diff
@@ -266,6 +278,30 @@ public String getProxyPassword() {
...
+    }
+
+    /**
+     * Sets whether access should be private or public
+     *
+     * @param access whether access is private or public
+     */
+    public void setAccess(String access) {
+        this.access = access;
```
smithkm: A null check before setting this would be a good idea.  Is there a reason you went with a String for this rather than an Enum?

geowebcache/s3storage/src/test/java/org/geowebcache/s3/S3BlobStoreConfigTest.java
smithkm: Thanks, much better.

geowebcache/s3storage/src/main/java/org/geowebcache/s3/S3BlobStoreConfig.java
engaric: Fair point.  I would often use an Enum in this kind of circumstance.  I guess the relative newness of working with xstream prevented me from thinking that way.  I would be prepared to work on that but am not sure of my allocation of work time this week.

PR #524, opened 2017-06-28
Fix WMSLayer proxyRequest null pointer exception

akashisama:
Check that content-type response header is not null before checking its value so proxy works with responses lacking a content-type header.


akashisama:
Apologies, the original incarnation of this PR broke a test, but I've fix it up now.

aaime:
You should actually try to write one (test) for this :-)

akashisama:
@aaime good point, test added

PR #525, opened 2017-07-13
Add functionality to cache vendor parameters when using GetCapibilities configuration

akashisama:
As described on the mailing list at https://sourceforge.net/p/geowebcache/mailman/message/35833768/


geowebcache/wms/src/main/java/org/geowebcache/config/GetCapabilitiesConfiguration.java
```diff
@@ -89,6 +90,8 @@
 
     private String vendorParameters = null;
 
+    private String cachedParameters = null;
```
smithkm: Maybe give it a collection of parameter filters instead of a String.  This fixes the problem with values containing `&` or `=` and with leaving all of the parameters completely open.

geowebcache/wms/src/test/java/org/geowebcache/config/GetCapabilitiesConfigurationTest.java
```diff
@@ -100,6 +105,18 @@ WebMapServer getWMS() {
         assertThat(wmsLayer.getLegends().getLegendsRawInfo().get(0).getWidth(), is(50));
         assertThat(wmsLayer.getLegends().getLegendsRawInfo().get(0).getHeight(), is(100));
         assertThat(wmsLayer.getLegends().getLegendsRawInfo().get(0).getFormat(), is("image/gif"));
+
+        List<ParameterFilter> parameterFilters = wmsLayer.getParameterFilters();
```
smithkm: Clearer and more compact way of doing this

````
assertThat(parameterFilters, containsInAnyOrder(
    hasProperty("key", equalToIgnoringCase("style")), 
    hasProperty("key", equalToIgnoringCase("angle"))));
````

I'd also like to see separate tests for empty and not empty.  The parsing also would also benefit from one, more than one, value with a `&` and value with a `=` and otherwise trying to break it.  Switching to a `Collection` would eliminate that need.

File: geowebcache/wms/src/main/java/org/geowebcache/config/GetCapabilitiesConfiguration.java [hunk 3]
```diff
@@ -270,6 +279,23 @@ private String getWMSUrl(WebMapServer wms) {
...
+                    String[] cparams = this.cachedParameters.split("&");
+                    for (String cp : cparams) {
+                        if (cp.length() > 0) {
+                            String[] split = cp.split("=", 2);
+                            String key = split[0];
+                            String defaultValue = split[1];
+                            log.info(key + " and " + defaultValue);
+                            if (key.length() > 0) {
+                                paramFilters.add(
+                                    new FreeStringParameterFilter(key, defaultValue)
```
smithkm: Completely free parameters are potentially dangerous. I try to discourage people from using them so having the GetCapabilitiesConfiguration using them behind the scenes strikes me as a bad idea.  I suppose we are already in that boat with the `NaiveWMSDimensionFilter` though.

geowebcache/wms/src/main/java/org/geowebcache/config/GetCapabilitiesConfiguration.java
```diff
@@ -270,6 +279,23 @@ private String getWMSUrl(WebMapServer wms) {
                     paramFilters.add(new NaiveWMSDimensionFilter(dimension, dimExtent));
                 }
 
+                if (this.cachedParameters != null && this.cachedParameters.length() != 0) {
+                    String[] cparams = this.cachedParameters.split("&");
+                    for (String cp : cparams) {
+                        if (cp.length() > 0) {
+                            String[] split = cp.split("=", 2);
```
smithkm: Need to handle `=` as part of a value.

geowebcache/wms/src/main/java/org/geowebcache/config/GetCapabilitiesConfiguration.java
```diff
@@ -270,6 +279,23 @@ private String getWMSUrl(WebMapServer wms) {
                     paramFilters.add(new NaiveWMSDimensionFilter(dimension, dimExtent));
                 }
 
+                if (this.cachedParameters != null && this.cachedParameters.length() != 0) {
+                    String[] cparams = this.cachedParameters.split("&");
```
smithkm: Need to handle `&` as part of a value.

akashisama:
Thanks for the review. I like the idea of sending a collection of filters to the constructor rather than a string, I'll give it a go.

akashisama:
I've made the code changes but I'm not sure how it should be documented. The use of GetCapabilitiesConfiguration is currently explained in the Quickstart section of the documentation, which "describes how to get quickly started with GeoWebCache without knowing much about the Configuration." I'm not sure the creation of parameter filters as java beans can be explained quickly. There's already a section in the documentation about creating Parameter Filters in the configuration xml but that's slightly different to the use of java beans.

Perhaps the best way to go is to send in a collection of pairs of key/default values and go back to creating the filters within GetCapabilitiesConfiguration.java? That way we avoid the previous difficulty of splitting up a single string, while also making it easy for new users.

smithkm:
I think there's a way to extend the Spring XML format with custom extensions so we could use the same format as in the configuration, but I expect that wouldn't play nicely with XStream.

Could you give an example of your split up configuration idea would like like? I'm not quite following you.

akashisama:
I'm thinking the configuration could be done like this:
```
<constructor-arg>
    <map>
        <entry key="parameterKey" value="defaultValue" />
        <entry key="parameterKey2" value="defaultValue" />
    </map>
</constructor-arg>
```

smithkm:
That's not as flexible as it only gives a default, I assume you intend an open "accept anything" filter?  We could always make other constructors in future if someone wants to do the work to make a more flexible option though so +1.

geowebcache/wms/src/test/java/org/geowebcache/config/GetCapabilitiesConfigurationTest.java
```diff
@@ -21,85 +29,203 @@
...
-        final WebMapServer server = createMock(WebMapServer.class);
-        WMSCapabilities cap = createMock(WMSCapabilities.class);
-        WMSRequest req = createMock(WMSRequest.class);
-        OperationType gcOpType = createMock(OperationType.class);
-        XMLConfiguration globalConfig = createMock(XMLConfiguration.class);
-        Capture<TileLayer> layerCapture = new Capture<TileLayer>();
-        
-        GetCapabilitiesConfiguration config = 
-                new GetCapabilitiesConfiguration(broker, url, mimeTypes, "3x3", "false"){
+    public class GetCapabilitiesConfigurationTest {
```
smithkm: You seem to have indented the entire class here, it's making it rather hard to evaluate the specific changes.

akashisama: The indentation fix is done in it's own commit so the other commits can be used individually to see relevant changes.

Issue #515, opened 2017-05-15
Extension point for per tile security


smithkm:
GeoServer is currently using a rather clunky mechanism to secure its embeded GWC instance.  An extension point that allows it to plug in a security module that can recognize individual layers and handle ROIs within them would allow for a much more maintainable and robust security system.  It could also be a starting point for improved data security within stand alone GWC.

Issue #319, opened 2015-03-24
Filtered Catalog


smithkm:
Allow the catalog of layers to be filtered.  Filters could hide, rename, or otherwise alter layers.

PR #526, opened 2017-08-09
#515 Data Security Framework

smithkm:
* Adds an extension point to secure access to data based on layer and geographic extent.
* Adds `GEOWEBCACHE_WMS_PROXY_REQUEST_WHITELIST` to control which request types are allowed to be proxied.
* Updates EasyMock to match the version in GeoServer and eliminates references Class Extensions since that is now in stock EasyMock.
* JUnit Rule to assist in testing of code that uses extensions.


smithkm:
Implements #515

geowebcache/kml/src/main/java/org/geowebcache/service/kml/KMLService.java
```diff
@@ -235,6 +239,14 @@ public void handleRequest(Conveyor conv) throws GeoWebCacheException {
             }
         }
         tile.setTileLayer(layer);
+        
+        if(Objects.nonNull(tile.getLayer())) {
+            if(tile.getTileIndex()[2]<0) {
```
tbarsballe: So, what does this mean? Either a comment here or some javadocs on ConveyorTile.getTileIndex() would be nice.

From what I can tell, looks like it might be checking the zoom level?

geowebcache/kml/src/main/java/org/geowebcache/service/kml/KMLService.java
```diff
@@ -235,6 +239,14 @@ public void handleRequest(Conveyor conv) throws GeoWebCacheException {
...
+                secDispatcher.checkSecurity(tile.getLayer(), null, null);
```
tbarsballe: As a follow-up to the above, why do we not need the bounding box / srs here?

Seems like it gets calculated by SecurityDispatcher if we are in the else clause

geowebcache/kml/src/main/java/org/geowebcache/service/kml/KMLService.java
```diff
@@ -467,7 +479,7 @@ private String getMimeTypeOverride(ConveyorKMLTile tile) {
      * @return
      * @throws ServiceException
      */
-    private static String createOverlay(ConveyorKMLTile tile, boolean isPackaged)
+    private String createOverlay(ConveyorKMLTile tile, boolean isPackaged)
```
tbarsballe: As an alternative to making this non-static, couldn't you just pass in the security dispatcher? (either way seems fine to me)

geowebcache/kml/src/main/java/org/geowebcache/service/kml/KMLService.java
```diff
@@ -707,4 +719,10 @@ private static double distance(double[] p1, double[] p2) {
         double dz = p1[2] - p2[2];
         return Math.sqrt(dx * dx + dy * dy + dz * dz);
     }
+
+    public void setSecurityDispatcher(SecurityDispatcher secDispatcher) {
+        this.secDispatcher = secDispatcher;
+    }
```
tbarsballe: For consistency, shouldn't there also be a `getSecurityDispatcher()` ?

geowebcache/kml/src/main/java/org/geowebcache/service/kml/KMLSiteMap.java
```diff
@@ -37,12 +38,14 @@
     private TileLayerDispatcher tLD = null;
     private StorageBroker storageBroker;
     private GridSetBroker gridSetBroker;
+    private SecurityDispatcher secDispatcher;
     
-    public KMLSiteMap(ConveyorKMLTile tile, TileLayerDispatcher tLD, GridSetBroker gridSetBroker) {
+    public KMLSiteMap(ConveyorKMLTile tile, TileLayerDispatcher tLD, GridSetBroker gridSetBroker, SecurityDispatcher secDispatcher) {
```
tbarsballe: GeoWebCacheDispatcher and KMLService both use setSecurityDispacher. This takes it in as an argument. Any reason for the difference?

geowebcache/kml/src/main/java/org/geowebcache/service/kml/KMZHelper.java
```diff
@@ -71,11 +72,13 @@
                         tileLayer.getName(), gridSetId, 
                         linkGridLocs[i], mime, null, null, null);
                 
-                tile.setTileLayer(tileLayer);
-                
                 // Apply request filters
                 try {
+                    tile.setTileLayer(tileLayer);
```
tbarsballe: Minor nitpick, but this doesn't really need to be inside the try block, only need the `secDisp.checkSecurity(tile);` here.

geowebcache/wms/src/main/java/org/geowebcache/service/wms/WMSService.java
```diff
@@ -287,20 +291,22 @@ public void handleRequest(Conveyor conv) throws GeoWebCacheException {
                 WMSGetCapabilities wmsCap = new WMSGetCapabilities(tld, tile.servletReq, servletBase, context, urlMangler);
```
tbarsballe: No security for getCapabilities?

geowebcache/wms/src/main/java/org/geowebcache/service/wms/WMSService.java
```diff
@@ -458,4 +474,37 @@ public void setHintsConfig(String hintsConfig) {
...
+    
+    protected void checkProxyRequest(String request) {
+        if(getProxyRequestWhitelist().stream().noneMatch(pattern-> pattern.equals("*")||pattern.equals(request))) {
+            throw new SecurityException("WMS Request "+request+" is not on request proxy whitelist");
+        }
+    }
+
+    public void setSecurityDispatcher(SecurityDispatcher securityDispatcher) {
+        this.securityDispatcher = securityDispatcher;
+    }
```
tbarsballe: Maybe add a "getSecurityDispatcher()" too?

geowebcache/wms/src/test/java/org/geowebcache/service/wms/WMSTileFuserTest.java
```diff
@@ -44,14 +47,17 @@
...
 import org.springframework.context.support.ClassPathXmlApplicationContext;
 import org.springframework.mock.web.MockHttpServletRequest;
 import org.springframework.mock.web.MockHttpServletResponse;
 
 public class WMSTileFuserTest extends TestCase {
     GridSetBroker gridSetBroker = new GridSetBroker(false, false);
+    SecurityDispatcher secDisp = mock(SecurityDispatcher.class);
     
     public void testTileFuserResolution() throws Exception {
+
```
tbarsballe: If you are adding the security dispatcher here, maybe also add one test for a security violation?

geowebcache/wmts/src/main/java/org/geowebcache/service/wmts/WMTSService.java
```diff
@@ -329,4 +333,8 @@ void addExtension(WMTSExtension extension) {
     public Collection<WMTSExtension> getExtensions() {
         return Collections.unmodifiableCollection(extensions);
     }
+
+    public void setSecurityDispatcher(SecurityDispatcher secDisp) {
+        this.securityDispatcher = secDisp;
+    }
```
tbarsballe: And `getSecurityDispatcher()`?

geowebcache/core/src/main/java/org/geowebcache/GeoWebCacheDispatcher.java
```diff
@@ -284,6 +288,10 @@ protected ModelAndView handleRequestInternal(HttpServletRequest request,
             OWSException owsE = (OWSException) e;
             writeFixedResponse(response, owsE.getResponseCode(), owsE.getContentType(),
                     owsE.getResponse(), CacheResult.OTHER);
+        } catch (SecurityException e) {
+            writeFixedResponse(response, 403, "text/plain",
+                    new ByteArrayResource("Not Authorized".getBytes()), CacheResult.OTHER);
```
tbarsballe: Do we just care about securing the layers, or also hiding their existence? I think GeoServer uses 404 for both secured layers and non-existing layers so you can't check if the layers exist based on the error response.

geowebcache/core/src/main/java/org/geowebcache/GeoWebCacheDispatcher.java
```diff
@@ -345,10 +353,21 @@ private void handleServiceRequest(String serviceStr, HttpServletRequest request,
...
+            if(conv instanceof ConveyorTile) {
+                ((ConveyorTile) conv).setTileLayer(layer);
+            }
+        } else {
+            layer = null;
         }
+        
 
         // Check where this should be dispatched
         if (conv.reqHandler == Conveyor.RequestHandler.SERVICE) {
```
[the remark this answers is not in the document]
smithkm: The services have to be responsible for security when doing their own thing.

geowebcache/wms/src/main/java/org/geowebcache/service/wms/WMSService.java
smithkm: That's certainly something we should do in the future but at the moment I'm just trying to get the existing behaviour done in a better way.

geowebcache/core/src/main/java/org/geowebcache/GeoWebCacheDispatcher.java
smithkm: This is why I added `GeoWebCacheException` as an option to be thrown by `checkSecurity`.  The filter can throw an exception for the tile being missing instead of a security one.

geowebcache/kml/src/main/java/org/geowebcache/service/kml/KMLService.java
smithkm: The KML request parser uses negative zoom levels as magic numbers.  I'll add a comment.

geowebcache/kml/src/main/java/org/geowebcache/service/kml/KMLService.java
smithkm: The problem is that the z index is a negative number which `boundsFromIndex` doesn't like.

geowebcache/kml/src/main/java/org/geowebcache/service/kml/KMLService.java
smithkm: I considered it but this felt more right.

geowebcache/kml/src/main/java/org/geowebcache/service/kml/KMLSiteMap.java
smithkm: For the Spring beans (Services etc) I started with properties, but this is instantiated normally which made it more convenient to make it a constructor argument.

geowebcache/kml/src/main/java/org/geowebcache/service/kml/KMZHelper.java
smithkm: I can't recall why I changed this so I've put it back.

geowebcache/wms/src/test/java/org/geowebcache/service/wms/WMSTileFuserTest.java
smithkm: Done

smithkm:
We already have one although it's not security specific https://github.com/GeoWebCache/geowebcache/issues/319

tbarsballe:
Ok, #319 seems fine then, especially since now that its been mentioned it will link back to this PR.

PR #531, opened 2017-08-28
Migrate GWC REST API to Spring MVC to keep in sync with GeoServer core

vickdw:
Resolves #532

PR #536, opened 2017-08-30
REST API Updates

tbarsballe:


jodygarnett:
Is this replaced by https://github.com/GeoWebCache/geowebcache/pull/531 ?

smithkm:
@jodygarnett This was meant to replace #531 but then we moved back to it.

PR #539, opened 2017-09-08
Fixing rest api to be backwards compatible.

vickdw:
Initial https://github.com/GeoWebCache/geowebcache/pull/531 split up the REST API into two:

* REST API for XML/JSON
* GWC GUI - implemented as REST API pages producing HTML content

This PR joins the two back together again to ease integration with GeoServer.

Splitting out the GWC user interface is a fine long term goal but will require coordination with downstream projects.


geowebcache/rest/src/main/java/org/geowebcache/rest/config/GWCRestConfig.java
```diff
@@ -0,0 +1,19 @@
+package org.geowebcache.rest.config;
```
jodygarnett: header

geowebcache/rest/src/main/java/org/geowebcache/rest/service/SeedService.java
```diff
@@ -0,0 +1,253 @@
+package org.geowebcache.rest.service;
```
jodygarnett: header

geowebcache/rest/src/main/java/org/geowebcache/rest/service/FormService.java
```diff
@@ -1,27 +1,4 @@
...
- * @author David Vick, Boundless, Copyright 2017
- *
- * Original file
- * SeedFormRestlet.java
- *
- */
-
-
-package org.geowebcache.rest.controller;
+package org.geowebcache.rest.service;
```
jodygarnett: header

jodygarnett:
think we should leave off merging this until integration with geoserver is sorted

geowebcache/rest/src/main/java/org/geowebcache/rest/config/GWCRestConfig.java
```diff
@@ -0,0 +1,35 @@
...
+/**
+ * Configure various aspects of Spring MVC
+ */
+@Configuration
+@EnableWebMvc
+public class GWCRestConfig extends WebMvcConfigurationSupport {
+
+    @Autowired
+    private ApplicationContext applicationContext;
+
```
tbarsballe: So, this class doesn't do anything right now... Any particular reason why it is here?

geowebcache/rest/src/main/java/org/geowebcache/rest/controller/GWCIndexController.java
```diff
@@ -35,7 +35,7 @@
 @RequestMapping(path="${gwc.context.suffix:}/rest")
 public class GWCIndexController {
 
-    @RequestMapping(value = "/about", method = RequestMethod.GET, produces = "text/html")
+    @RequestMapping(method = RequestMethod.GET, produces = "text/html")
```
tbarsballe: I assume this is the root of the rest API ? Does it also need to be available from `/about`, in which case we would want `value = {"", "/about"}` ?

geowebcache/web/src/main/webapp/WEB-INF/geowebcache-rest-context.xml
```diff
@@ -7,7 +7,6 @@
     http://www.springframework.org/schema/context http://www.springframework.org/schema/context/spring-context-4.0.xsd
     http://www.springframework.org/schema/mvc http://www.springframework.org/schema/mvc/spring-mvc-4.0.xsd">
 
-  <mvc:annotation-driven />
```
tbarsballe: How is anything working without this ?!

geowebcache/rest/src/main/java/org/geowebcache/rest/config/GWCRestConfig.java
```diff
@@ -0,0 +1,35 @@
...
+import org.springframework.context.annotation.Configuration;
+import org.springframework.web.servlet.config.annotation.EnableWebMvc;
+import org.springframework.web.servlet.config.annotation.WebMvcConfigurationSupport;
+
+/**
+ * Configure various aspects of Spring MVC
+ */
+@Configuration
+@EnableWebMvc
+public class GWCRestConfig extends WebMvcConfigurationSupport {
```
tbarsballe: So, this class doesn't do anything right now... Any particular reason why it is here?

geowebcache/rest/src/main/java/org/geowebcache/rest/controller/GWCIndexController.java
vickdw: "About" was a bad endpoint, the former IndexRestlet responded on the /rest endpoint. This may also be contributing to the GS root rest endpoint not working.

geowebcache/web/src/main/webapp/WEB-INF/geowebcache-rest-context.xml
vickdw: GeoServer is in the same situation, the rest module doesn't have this tag in it's context.xml.

geowebcache/rest/src/main/java/org/geowebcache/rest/config/GWCRestConfig.java
vickdw: No, this class can be removed at this point.

geowebcache/rest/src/main/java/org/geowebcache/rest/config/GWCRestConfig.java
vickdw: Done.

geowebcache/rest/src/main/java/org/geowebcache/rest/service/SeedService.java
vickdw: Done.

geowebcache/rest/src/main/java/org/geowebcache/rest/config/GWCRestConfig.java
tbarsballe: So, after talking with @dvntucker , looks like the `@EnableWebMvc` here is doing the same thing that `<mvc:annotation-driven />` was, so when we remove this we will need to re-add that to the `geowebcache-rest-context.xml` (GeoServer has a Config class with the `@EnableWebMvc` annotation)
Since this is in WEB-INF, it may make GWC play nicer with the GS config.

geowebcache/rest/src/main/java/org/geowebcache/rest/config/GWCRestConfig.java
```diff
@@ -0,0 +1,19 @@
...
+
+    @Autowired
+    private ApplicationContext applicationContext;
+
+}
```
dvntucker: I don't think this class should be needed at all? If there is something that's needed from this, it would be best if it could be moved to geowebcache-rest-context.xml, otherwise it will interfere with GeoServer rest config.

geowebcache/web/src/main/webapp/WEB-INF/geowebcache-rest-context.xml
dvntucker: This should be left in. This should be the place where MVC is enabled, since we know it's not used by GeoServer, only GWC standalone.

tbarsballe:
All feedback has now been addressed, plus a few other fixes. CI passes, this looks ready to merge.

PR #540, opened 2017-09-12
fixed bug in reload controller and added a missing test for the contr…

vickdw:
…oller.

### DIFF
--- a/geowebcache/rest/src/main/java/org/geowebcache/rest/controller/ReloadController.java
+++ b/geowebcache/rest/src/main/java/org/geowebcache/rest/controller/ReloadController.java
@@ -22,6 +22,7 @@
 
 package org.geowebcache.rest.controller;
 
+import com.google.common.base.Splitter;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 import org.geowebcache.GeoWebCacheException;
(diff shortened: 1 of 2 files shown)


geowebcache/rest/src/main/java/org/geowebcache/rest/controller/ReloadController.java
```diff
@@ -54,14 +57,30 @@
 
     @RequestMapping(value = "/reload", method = RequestMethod.POST)
```
tbarsballe: We should maybe have an `accepts=application/x-www-form-urlencoded` here, but since this is the only implementation for this endpoint I think it is fine as-is.

geowebcache/rest/src/test/java/org/geowebcache/rest/reload/ReloadTest.java
```diff
@@ -0,0 +1,78 @@
...
+        "file*:/webapp/WEB-INF/geowebcache-servlet.xml"
+})
+public class ReloadTest {
+    private MockMvc mockMvc;
+    ReloadController reload;
+
+    TileLayerDispatcher tld;
+
+    @Before
+    public void setup() throws GeoWebCacheException {
```
tbarsballe: Thanks for the test case

PR #542, opened 2017-09-14
fixing a bug in GWC when GWC was embedded in GS causing the contents …

vickdw:
…of the HttpServletRequest to have already been streamed before the Controller recieved the request.


vickdw:
I've changed the names for consistency.  To get the code working in standalone I removed unnecessary configuration overrides, these same configuration overrides are required within the context of GeoServer.  The edits made in this PR allow GWC to behave consistently in both Standalone and Embedded modes.  Debugging should not be affected, and should be the same across both modes.

I could not identify a good test case for building a regression test for these changes. I changed controller by adding an Annotation for the RequestBody to be handled as a String instead of an InputStream or being read by a BufferedReader and changed the services to accept the String.

PR #544, opened 2017-09-29
bug fixes where clicking submit on Seeding form returned blank page and leaving the BBOX of the Seeding form blank caused an error. GEOS-8312 and GEOS-8313

vickdw:
Fixes the blank page that is returned when clicking submit on the seeding form and also fixes the error received when clicking submit while leaving the BBOX entries blank.

### DIFF
--- a/geowebcache/rest/src/main/java/org/geowebcache/rest/service/FormService.java
+++ b/geowebcache/rest/src/main/java/org/geowebcache/rest/service/FormService.java
@@ -182,27 +182,26 @@ public class FormService {
             if (tl == null) {
                 throw new RestException("No layer specified", HttpStatus.BAD_REQUEST);
             }
-            handleDoSeedPost(params, tl);
-        } else {
-            throw new RestException(
-                    "Unknown or malformed request. Please try again, somtimes the form "
-                            + "is not properly received. This frequently happens on the first POST "
-                            + "after a restart. The POST was to " + request.getRequestURI(),
-                    HttpStatus.BAD_REQUEST);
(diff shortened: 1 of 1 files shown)


geowebcache/rest/src/main/java/org/geowebcache/rest/service/FormService.java
```diff
@@ -182,28 +182,31 @@
...
     }
 
     private ResponseEntity<?> handleDoSeedPost(Map<String, String> form, TileLayer tl) throws RestException,
             GeoWebCacheException {
+        String gridSetId = form.get("gridSetId");
+
         BoundingBox bounds = null;
 
-        if (form.get("minX") != null) {
+        if (!form.get("minX").equalsIgnoreCase("")) {
```
tbarsballe: `form.get("minX") != null && !form.get("minX").equalsIgnoreCase("")` is probably safer

geowebcache/rest/src/main/java/org/geowebcache/rest/service/FormService.java
```diff
@@ -182,28 +182,31 @@
             if (tl == null) {
                 throw new RestException("No layer specified", HttpStatus.BAD_REQUEST);
             }
-            handleDoSeedPost(params, tl);
+            return handleDoSeedPost(params, tl);
         } else {
```
tbarsballe: Technically, the else is now unnecessary. Probably clearer to just remove it and have the exception at top level.

tbarsballe:
Backported to 1.12.x

PR #545, opened 2017-10-02
Add test for REST request filter update endpoint

tbarsballe:


geowebcache/rest/src/main/java/org/geowebcache/rest/config/XMLConfiguration.java
```diff
@@ -55,9 +55,6 @@ private static XStream getConfiguredXStream(XStream xs) {
         xs.alias("gwcConfiguration", GeoWebCacheConfiguration.class);
         xs.useAttributeFor(GeoWebCacheConfiguration.class, "xmlns_xsi");
         xs.aliasField("xmlns:xsi", GeoWebCacheConfiguration.class, "xmlns_xsi");
-        xs.useAttributeFor(GeoWebCacheConfiguration.class, "xsi_noNamespaceSchemaLocation");
-        xs.aliasField("xsi:noNamespaceSchemaLocation", GeoWebCacheConfiguration.class,
-                "xsi_noNamespaceSchemaLocation");
```
tbarsballe: This appears to have been removed from GeoWebCacheConfiguration by this commit: https://github.com/GeoWebCache/geowebcache/commit/2bc483bba56bc02377c5e8ef6802f8f2b2f84701#diff-984a488a982efb0f04c846b1ea5a0b5bL39 , but never removed from XMLConfiguration.

This inconsistency causes the following error when POSTing to the filter update endpoint:
```
Caused by: com.thoughtworks.xstream.converters.reflection.MissingFieldException: No field 'xsi_noNamespaceSchemaLocation' found in class 'org.geowebcache.config.GeoWebCacheConfiguration'
```

It is likely that this endpoint has been nonfunctional since the above commit.

PR #547, opened 2017-10-04
Serve OpenLayers static resources through WebResource controller.

### LINKED ISSUE #541: Demo pages failing to load OL static resources under /geowebcache context path
I've been working off of the master branch of GWC for some testing, and some commits in the past few weeks seem to have disabled the demo pages (the links to static resources are failing to include the /geowebcache path in their URLs, returning 404 errors).

For ex:
```
http://localhost:8090/geowebcache/demo/layer01?gridSet=EPSG:4326&format=image/png

Request URL:http://localhost:8090/openlayers3/ol.js
Request Method:GET
Status Code:404 Not Found
Remote Address:[::1]:8090
Referrer Policy:no-referrer-when-downgrade

Request URL:http://localhost:8090/openlayers3/ol.css
Request Method:GET
Status Code:404 Not Found
Remote Address:[::1]:8090
Referrer Policy:no-referrer-when-downgrade
```

I don't know the code too well, but I see a bunch of possibly related changes here:  https://github.com/GeoWebCache/geowebcache/commit/bb5ce18eecb5a7a027bd1807a7e629de298bb2e1.

Are there any tests that would find this?  `mvn clean install` worked for me, so I don't think it's a configuration/environment issue.

smithkm:
Fixes #541 error loading demo page resulting from switch to Spring MVC.


geowebcache/rest/src/test/resources/org/geowebcache/rest/webresources/test.js
```diff
@@ -0,0 +1 @@
+JS
```
tbarsballe: This seems strange, does it really need to have content?

geowebcache/core/src/main/resources/org/geowebcache/rest/webresources/openlayers3/ol.js
```diff
@@ -0,0 +1,998 @@
+// OpenLayers 3. See http://openlayers.org/
```
tbarsballe: This is marked as a new file, and I can't find a corresponding deleted file.

Elsewhere, I see 
```
...che/web/src/main/webapp/openlayers3/ol.js → ...s/org/geowebcache/rest/webresources/ol.js
```

So it looks like there might be a duplicate ol3 file?

geowebcache/rest/src/test/resources/org/geowebcache/rest/webresources/test.js
smithkm: Not really but the others had it so I included it for consistency.  I think when I originally created these tests I was originally planning to check that the content was included but never added that assertion.

geowebcache/core/src/main/resources/org/geowebcache/rest/webresources/openlayers3/ol.js
smithkm: Ah I made a duplicate when I ran into trouble with subdirectories and forgot to delete it, this is the correct one, it's the one that did get marked as a rename that shouldn't be there.  Thanks for catching that.

geowebcache/rest/src/test/resources/org/geowebcache/rest/webresources/test.js
tbarsballe: Fair enough

PR #559, opened 2017-11-18
Fix NPE when checking security of tile fuser requests, Fixes #558

### LINKED ISSUE #558: TileFuser does not set tileLayer on conveyor
When constructing a ConveyorTile, TileFuser does not set the tileLayer, which can lead to an NPE when a SecurityFilter checks it.

https://sourceforge.net/p/geoserver/mailman/geoserver-users/thread/CALi7temUxpSfn-iN1CVxyOfs3cWQpNQQzVPowtrFvGfwat%3DmYw%40mail.gmail.com/#msg36108733

smithkm:
Sets the `tileLayer` field on the tile requests made by the `TileFuser` so that the security check doesn't give an NPE.  See #558

### DIFF
--- a/geowebcache/wms/src/main/java/org/geowebcache/service/wms/WMSTileFuser.java
+++ b/geowebcache/wms/src/main/java/org/geowebcache/service/wms/WMSTileFuser.java
@@ -308,7 +308,8 @@ public class WMSTileFuser{
             hints = HintsLevel.getHintsForMode(values.get("hints")).getRenderingHints();
         }        
     }
-
+    
+    @Deprecated
     protected WMSTileFuser(TileLayer layer, GridSubset gridSubset, BoundingBox bounds, int width,
             int height) {
         this.sb = null;
@@ -526,6 +527,8 @@ public class WMSTileFuser{
 
                 ConveyorTile tile = new ConveyorTile(sb, layer.getName(), gridSubset.getName(),
                         gridLoc, srcFormat, fullParameters, null, null);
+                
+                tile.setTileLayer(layer);
                 
                 securityDispatcher.checkSecurity(tile);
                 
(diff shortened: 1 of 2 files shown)


geowebcache/wms/src/main/java/org/geowebcache/service/wms/WMSTileFuser.java
```diff
@@ -308,7 +308,8 @@ protected WMSTileFuser(TileLayerDispatcher tld, StorageBroker sb, HttpServletReq
             hints = HintsLevel.getHintsForMode(values.get("hints")).getRenderingHints();
         }        
     }
-
+    
+    @Deprecated
```
tbarsballe: Javadoc indicating what to use instead would be helpful

geowebcache/wms/src/test/java/org/geowebcache/service/wms/WMSTileFuserTest.java
```diff
@@ -78,23 +96,26 @@ public void testTileFuserResolution() throws Exception {
...
+        WMSTileFuser tileFuser = new WMSTileFuser(tld, sb, fuserRequest(layer, gridSubset, bounds, width, height));
         tileFuser.setSecurityDispatcher(secDisp);
         tileFuser.determineSourceResolution();
         assertEquals(0.087890625, tileFuser.srcResolution, 0.087890625*0.001);
         
         // Zoomed too far out
         height = (int) bounds.getWidth() / 10;
         width = (int) bounds.getWidth() / 10;
-        tileFuser = new WMSTileFuser(layer, gridSubset, bounds, width, height);
+        tileFuser = new WMSTileFuser(tld, sb, fuserRequest(layer, gridSubset, bounds, width, height));
```
tbarsballe: With these changes, this no longer tests the `@Deprecated` constructor. 

If the constructor is still used anywhere in GWC, a simple test using that constructor is probably a good idea, given the amount of setup that seems to be done in it. If that constructor is not used anywhere anymore, seems fine as is.

smithkm: The constructor appeared to exist only for the sake of the tests, resulting in different behaviour between what was tested and what happens in real usage.  I considered removing the constructor entirely bu deprecated to play it safe although if there's anything using it out there it will likely fall afoul of NPEs in the security filter.

tbarsballe: Ok, in that case everything looks good.

PR #560, opened 2017-11-20
[GEOS-8401] Fix SeedController Post for embedded GWC

tbarsballe:
See https://osgeo-org.atlassian.net/browse/GEOS-8401

Only seems to affect GWC embedded in GeoServer. As such, test cases are downstream in GeoServer: https://github.com/geoserver/geoserver/pull/2652


tbarsballe:
Backported to 1.12.x

Issue #564, opened 2017-11-22
GeoWebCache uses application/vnd.ogc.wms_xml mime type for WMTS capabilities document


nmco:
According to WMTS 1.0 OGC CITE tests the mime type of WMTS capabilities document should be a mime type that ends in ``\xml``.

Issue #563, opened 2017-11-22
Makes LegendBuilder correctly handle extreme scales values


nmco:
Related GeoServer pull request: https://github.com/geoserver/geoserver/pull/2639

nmco:
GeoServer is a bit weird it uses the extremes values as default min and max scales values instead of ``NULL``, which means that any code who need to handle these values needs to have a special check for them. 

In GWC we can handle this in a single point by making the ``LegendBuilder`` consider this values as ``NULL``.

PR #565, opened 2017-11-22
Fix issues found during WMTS 1.0 CITE tests

nmco:
Associates issues:
* https://github.com/GeoWebCache/geowebcache/issues/563
* https://github.com/GeoWebCache/geowebcache/issues/564

This PR allows Geoserver to compute MinScales and MaxSscales for styles legends. It also change the mime type used for WMTS capabilities documents to ``text\xml``.


nmco:
I agree with you when one of the extremes is set we should set the value to ``NULL``, good catch. 

Regarding the issue name, GeoServer is a bit weird it uses the extremes values as default instead of ``NULL`` which means that any code who need to handle \ use those values needs to have a special check for them ... but yes I agree the issue name could be more explicit.

nmco:
``LegendInfoBuilder`` updated to set ``NULL`` values instead of just ignoring them. The associated was also updated with better name and description. Thanks for the review @smithkm !

PR #569, opened 2017-11-28
Fixes #568: support for parameters from ENV in S3BlobStore configuration

### LINKED ISSUE #568: Introduce ENV parameters support for the S3BlobStore
We would like to make S3 connection parametric (e.g. bucket, prefix, keys, etc.)

We could use the work done here: https://github.com/GeoWebCache/geowebcache/pull/399

mbarto:


geowebcache/s3storage/src/test/java/org/geowebcache/s3/S3BlobStoreConfigProviderTest.java
```diff
@@ -0,0 +1,65 @@
...
+        ClassPathXmlApplicationContext context = new ClassPathXmlApplicationContext(
+                "appContextTestS3.xml");
+
+        GeoWebCacheExtensions gse = new GeoWebCacheExtensions();
+        gse.setApplicationContext(context);
+    }
+
+    @After
+    public void tearDown() throws Exception {
+        System.setProperty("BUCKET", "");
```
aaime: This sets the system property to the empty strings but does not remove it. To make extra sure to leave nothing behind, best to use [clearProperty](https://docs.oracle.com/javase/7/docs/api/java/lang/System.html#clearProperty(java.lang.String)) instead.

geowebcache/s3storage/src/main/java/org/geowebcache/s3/S3BlobStoreConfigProvider.java
```diff
@@ -41,20 +48,53 @@ public Object fromString(String str) {
...
         public Object fromString(String str) {
+            str = resolveFromEnv(str);
             if (Strings.isNullOrEmpty(str)) {
                 return null;
             }
             return super.fromString(str);
         }
     };
 
+    private static SingleValueConverter NullableStringConverter = new StringConverter() {
```
aaime: This is not just a "nullable" converter, it's more of a "EnvironmentStringConverter", it's main job is to pull info from the environment. As coded it might also unwillingly pick values that were not meant to be variable, as it's "always on".
In GeoServer we have a flag to control parameter injection from the environment (ALLOW_ENV_PARAMETRIZATION) and GeoWebCache has the same here: [GeoWebCacheEnvironment](https://github.com/GeoWebCache/geowebcache/blob/926e5e8ded827d35b3cbfd353af36387cde50f29/geowebcache/core/src/main/java/org/geowebcache/GeoWebCacheEnvironment.java#L45)
Imho that should be used, following an approach similar to the existing ones for uniformity:
https://github.com/GeoWebCache/geowebcache/blob/73afa58a2154d2ef0e223812cf8b8569e1b9d51b/geowebcache/diskquota/core/src/main/java/org/geowebcache/diskquota/ConfigLoader.java#L292

aaime: Sorry, did not see the "resolveFromEnv" code, my apologies!
The name could still use some love though :-)

aaime:
Build failed on the usual REST online test that randomly fails... I've kicked it again.

PR #577, opened 2017-12-12
[GEOS-8471] Fix 406 error on embedded GWC seed form POST

tbarsballe:
See https://osgeo-org.atlassian.net/browse/GEOS-8471

When the content-type of a POST request is `application/x-www-form-urlencoded`, the BufferedRequestWrapper will add the POST content to the query parameters of the request.

Spring MVC has an option where the `format` parameter can be used to determine the content-type. This is disabled by default, but is enabled in GeoServer (consequently, the issue only affects embedded GWC).

These two features were causing the format parameter of the seed POST to be interpreted as the requested response type (e.g. `image/png`), which was not a valid type for the GWC Rest API (only supports html, json and xml), resulting in the 406 response.

Since this only affects embedded GWC, test cases are in GeoServer: https://github.com/geoserver/geoserver/pull/2683

### DIFF
--- a/geowebcache/rest/src/main/java/org/geowebcache/rest/service/FormService.java
+++ b/geowebcache/rest/src/main/java/org/geowebcache/rest/service/FormService.java
@@ -206,7 +206,7 @@ public class FormService {
         int threadCount = Integer.parseInt(form.get("threadCount"));
         int zoomStart = Integer.parseInt(form.get("zoomStart"));
         int zoomStop = Integer.parseInt(form.get("zoomStop"));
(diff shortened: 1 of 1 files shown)


tbarsballe:
Ultimately it is a consequence of having embedded GWC use the same MVC context as the rest of GeoServer (I'm not sure there is any other way to do this that is remotely sensible, but if we wanted to look into separating them somehow that would prevent future issues of this form).

tbarsballe:
Backported to 1.12.x

PR #578, opened 2017-12-14
[GEOS-8474] Fix 'Chunk [] is not a valid entry' error in embedded GWC seed form

tbarsballe:
See https://osgeo-org.atlassian.net/browse/GEOS-8474

Closely related to https://github.com/GeoWebCache/geowebcache/pull/577 , but only occurs when the GeoServer LoggingFilter is disabled (it is disabled by default in 2.12, but enabled by default in 2.13)

Occurs when the request InputStream is empty / already read / EOF. This occurs only when Content-Type is `application/x-www-urlencoded` (i.e., POST from the seed form page). Enabling the LoggingFilter wraps the request in a BufferedRequestWriter that preserves the contents of the InputStream.

Fixed by getting the urlencoded parameters from the parameter map. Still need to read the body as well, in order to support cases when the Content-Type is not `application/x-www-urlencoded` (which GeoWebCache seems to support)

### DIFF
--- a/geowebcache/rest/src/main/java/org/geowebcache/rest/controller/SeedController.java
+++ b/geowebcache/rest/src/main/java/org/geowebcache/rest/controller/SeedController.java
@@ -23,6 +23,7 @@
 
 package org.geowebcache.rest.controller;
 
+import com.google.common.base.Splitter;
 import org.geowebcache.GeoWebCacheException;
 import org.geowebcache.config.XMLConfiguration;
 import org.geowebcache.rest.exception.RestException;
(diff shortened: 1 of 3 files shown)


tbarsballe:
Backported to 1.12.x

eva407:
Hi @tbarsballe how do I enable GeoServer LoggingFilter to be able to seed my layer. I cant see this option in admin page, is it done by changing a file on server disk?

tbarsballe:
@eva407 - You can enable the LoggingFilter by editing `WEB-INF/web.xml` in your GeoServer webapp, changing the value of the enabled parameter of the "Request Logging Filter" to `true` (should be around line 116 in that file).

eva407:
@tbarsballe thanks so much for a prompt response, unfortunately it still is not working after re-starting the server, any other ideas why?

tbarsballe:
@eva407 - Do you have any extensions installed? Some extensions override the logging filter settings (hazelcast clustering is the only one I know of). If it is still not working, you are probably best off either using a nightly build or waiting for the 2.12.2 release later this month.

eva407:
We do not have any extensions installed. Thank you for your help. We will watch for 2.12.2.

PR #583, opened 2017-12-22
Fix servlet context path parsing

master-passeli:
This patch fixes servlet context path parsing which fails in cases where
HTTP request Host attribute does not match exactly with request URL

Example:
geowebcache running in localhost:8080 
`$ curl -i -H 'Host: localhost:80' "http://localhost:8080/service/wms?SERVICE=WMS&VERSION=1.1.1&REQUEST=getcapabilities&TILED=true"`
 results getCapabilities result with lines:

`<OnlineResource xmlns:xlink="http://www.w3.org/1999/xlink" xlink:type="simple" xlink:href="http://localhost/service/wms?SERVICE=WMS&amp;"/>`

This is fine but...

`$ curl -i -H 'Host: localhost:443' "http://localhost:8080/service/wms?SERVICE=WMS&VERSION=1.1.1&REQUEST=getcapabilities&TILED=true"`

Produces invalid result

`<OnlineResource xmlns:xlink="http://www.w3.org/1999/xlink" xlink:type="simple" xlink:href="http://localhost/:443/service/wms?SERVICE=WMS&amp;"/>`

Patch fixes case where ServletUtils.getServletContextPath(HttpServletRequest, String, String) method refers ServletUtils.getServletBaseURL(HttpServletRequest, String) when finding out the context. Latter method is constructing URL based on HTTP's requests "Host" attribute which might differ from URL requested (like in example). Some load balancers does that. Patch modfies the way how those methods produces they result. 

Addition to thet, this patch also adds some validation for Host attribute and request URL which are
later used in generating getCapabilities response. Validation should prevent injecting arbitary text in that response.

Example: 
`$ curl -i -H 'Host: nasty_code%we_dont_want_to_end_up_anywhere!><magic strings>' "http://localhost:8080/service/wms?SERVICE=WMS&VERSION=1.1.1&REQUEST=getcapabilities&TILED=true"`

`<OnlineResource xmlns:xlink="http://www.w3.org/1999/xlink" xlink:type="simple" xlink:href="http://nasty_code%we_dont_want_to_end_up_anywhere!&gt;&lt;magic strings&gt;/service/wms?SERVICE=WMS&amp;"/>`

I'm not sure this is actually a security problem but old behaviour is not valid for sure.

### DIFF
--- a/geowebcache/core/src/main/java/org/geowebcache/util/ServletUtils.java
+++ b/geowebcache/core/src/main/java/org/geowebcache/util/ServletUtils.java
@@ -19,6 +19,8 @@ package org.geowebcache.util;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.UnsupportedEncodingException;
+import java.net.URI;
+import java.net.URISyntaxException;
 import java.net.URLDecoder;
 import java.net.URLEncoder;
 import java.text.ParseException;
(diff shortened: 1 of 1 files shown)


smithkm:
Giving Travis a poke to see if it passes.

master-passeli:
That Travis build does not seems to work. Log says build stalled or something even all the unit tests did run ok. This patch should definitely build ok.

smithkm:
All the tests up to the point where it stalled passed but the tests that would have been run after that never got a chance to run.  I also noticed that you are pulling against the `1.12.x` branch.  Our policy is to always start on `master` then backport except in cases of a bug specific to a branch.

master-passeli:
Closing this pull request and sent a new one ported to Master. New request #618

Issue #621, opened 2018-03-14
Request proxying doesn't work in 1.13-RC1


jonescc:
In 1.11.4, the following request is proxied through to the associated server

    http://localhost:8080/geowebcache/service/wms?request=enabledFilters&service=layerFilters&version=1.0.0&layer=imos:argo_primary_profile_core_low_res_good_qc_data

In 1.13-RC1, this doesn't work - I get a null pointer exception as follows:

```
14 Mar 11:23:45 (http-bio-8080-exec-6) ERROR [geowebcache.GeoWebCacheDispatcher] - Request failed
java.lang.NullPointerException
	at java.util.Hashtable.get(Hashtable.java:364)
	at org.geowebcache.layer.AbstractTileLayer.getGridSubset(AbstractTileLayer.java:511)
	at org.geowebcache.conveyor.ConveyorTile.getGridSubset(ConveyorTile.java:179)
	at org.geowebcache.filter.security.SecurityDispatcher.checkSecurity(SecurityDispatcher.java:59)
	at org.geowebcache.service.wms.WMSService.handleRequest(WMSService.java:308)
	at org.geowebcache.GeoWebCacheDispatcher.handleServiceRequest(GeoWebCacheDispatcher.java:394)
	at org.geowebcache.GeoWebCacheDispatcher.handleRequestInternal(GeoWebCacheDispatcher.java:272)
	at org.springframework.web.servlet.mvc.AbstractController.handleRequest(AbstractController.java:147)
	at org.geowebcache.controller.GeoWebCacheDispatcherController.handleRestApiRequest(GeoWebCacheDispatcherController.java:31)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
   ...
```

smithkm:
I'm unable to replicate this.  Proxied requests are working fine for me.

As best I can tell this error would be a result of the `initialize(GridSetBroker gridSetBroker)` method not having been called on the layer.

Have you messed around with the Spring context at all?

smithkm:
[...]

PR #623, opened 2018-03-15
Null check gridsetId in ConveyorTile before getting grid subset

smithkm:
See #621


smithkm:
This was introduced int he 1.12 branch so needs to be backported to both 1.13 and 1.12

tbarsballe:
Answering my own questions:
1. You could perhaps add the null check to AbstractTileLayer instead. The behaviour for a null id is undefined, although throwing an NPE does seem to be a reasonable behaviour at that level, so I think the current fix is fine.
2. Yes, SecurityDispatcher checks if the returned subset is null

Issue #620, opened 2018-03-13
Can't override default 4326 settings in 1.13-RC1


jonescc:
In 1.11.4, we override the default set of resolutions for 4326 by redefining the default 4326 gridset in geowebcache.xml as follows:

 <gridSets>

    <gridSet>
      <name>EPSG:4326</name>
      <description>A default WGS84 tile matrix set where the first zoom level covers the world with two tiles
        on the horizonal axis and one tile over the vertical axis and each subsequent zoom level is calculated
        by half the resolution of its previous one.
      </description>
      <srs>
        <number>4326</number>
      </srs>
      <extent>
        <coords>
          <double>-180.0</double>
          <double>-90.0</double>
          <double>180.0</double>
          <double>90.0</double>
        </coords>
      </extent>
      <alignTopLeft>false</alignTopLeft>
      <resolutions>
        <double>1.40625</double>   
        <double>0.703125</double>
        <double>0.3515625</double>   
        <double>0.17578125</double>
        <double>0.087890625</double>
        <double>0.0439453125</double>
        <double>0.02197265625</double>
        <double>0.010986328125</double>
        <double>0.0054931640625</double>
        <double>0.00274658203125</double>
        <double>0.001373291015625</double>
        <double>6.866455078125E-4</double>
        <double>3.433227539062E-4</double>
        <double>1.716613769531E-4</double> 
      </resolutions>
      <metersPerUnit>111319.49079327358</metersPerUnit>
      <pixelSize>2.8E-4</pixelSize>

      <tileHeight>256</tileHeight>
      <tileWidth>256</tileWidth>

      <yCoordinateFirst>false</yCoordinateFirst>
    </gridSet>
  </gridSets>

Extract from 1.11.4 log where override definition is recognised and replaces default.

```
14 Mar 09:23:31 (RMI TCP Connection(2)-127.0.0.1) INFO [config.XMLFileResourceProvider] - Found configuration file in /home/craigj/tomcat7/apache-tomcat-7.0.70/temp/geowebcache
14 Mar 09:23:31 (RMI TCP Connection(2)-127.0.0.1) INFO [config.XMLConfiguration] - Configuration file validated fine.
[...]

PR #626, opened 2018-03-16
Change DefaultGridSet priority

emerkle826:
This change allows any of the internal default GridSets to be overridden in geowebcache.xml. 

Addresses https://github.com/GeoWebCache/geowebcache/issues/620

### DIFF
--- a/geowebcache/core/src/main/java/org/geowebcache/grid/GridSetBroker.java
+++ b/geowebcache/core/src/main/java/org/geowebcache/grid/GridSetBroker.java
@@ -183,11 +183,33 @@ public class GridSetBroker implements ConfigurationAggregator<GridSetConfigurati
     }
     
     public GridSet getWorldEpsg4326() {
-        return getDefaults().worldEpsg4326();
+        // get the internal default. We need it's name, if nothing else.
+        final GridSet internalDefault4326GridSet = getDefaults().worldEpsg4326();
+        // get the name of the internal default as it may not always be "EPSG:4326"
+        final String internalDefault4326Name = internalDefault4326GridSet.getName();
+        // Use the highest priority EPSG:4326 gridset we have (either an overriden one, or the defaults)
+        Optional<GridSet> gridSet = getGridSet(internalDefault4326Name);
+        if (gridSet.isPresent()) {
+            return gridSet.get();
+        }
+        // probably won't ever hit this, but if 4326 isn't in the configuration at this point, just return the internal
+        // defaults.
+        return internalDefault4326GridSet;
     }
 
     public GridSet getWorldEpsg3857() {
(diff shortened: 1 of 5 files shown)


geowebcache/core/src/test/java/org/geowebcache/config/XMLConfigurationTest.java
```diff
@@ -369,6 +373,51 @@ public void testSaveGridSet() throws Exception {
...
+        GeoWebCacheExtensions gwcExtensions = new GeoWebCacheExtensions();
+        gwcExtensions.setApplicationContext(appContext);
+        // get the GridSet for 4326, should be the override in the test XML file
+        GridSet override4326 = gridSetBroker.get("EPSG:4326");
+        assertNotNull(override4326);
+        // get the default GridSet for 4326.
+        GridSet worldEpsg4326 = defaultGridSets.worldEpsg4326();
+        // override should have a different resolution list
+        assertEquals("Unexpected number of Default EPSG:4326 reslution levels", 22, worldEpsg4326.getNumLevels());
+        assertEquals("Unexpected number of Overriden EPSG:4326 reslution levels", 14, override4326.getNumLevels());
```
tbarsballe: Minor typo on these to lines - _resolution_

tbarsballe:
Update - I took a look at where those methods get called, and it looks like they only go through the GridSetBroker - `getWorldEpsg4326()` and `getWorldEpsg3857()`, so it should be a relatively easy fix.

geowebcache/core/src/main/java/org/geowebcache/grid/GridSetBroker.java
```diff
@@ -183,10 +183,24 @@ public DefaultGridsets getDefaults() {
     }
     
     public GridSet getWorldEpsg4326() {
+        // Use the highest priority EPSG:4326 gridset we have (either an overriden one, or the defaults)
+        Optional<GridSet> gridSet = getGridSet("EPSG:4326");
```
tbarsballe: If you look in DefaultGridsets, you'll see that the name of the default EPSG:4326 gridset is not always "EPSG:4326": https://github.com/GeoWebCache/geowebcache/blob/master/geowebcache/core/src/main/java/org/geowebcache/config/DefaultGridsets.java#L58-L69

As I noted in my initial review, you'll need to get the default names from the DefaultGridsets object

So something like:
```
getGridSet(getDefaults().worldEpsg4326().getName())
```

emerkle826: I see now. One more swing in a bit.

tbarsballe:
Backported to 1.13.x

Issue #640, opened 2018-04-16
WMTS operations URLs are invalid with AuthKey module 


ianturton:
If the AuthKey module is in use it adds a parameter to the baseURL and then GeoWebCache adds a ? to the end of the URL when it should use & instead. 

The issue is in line 390 of /gwc-wmts/src/main/java/org/geowebcache/service/wmts/WMTSGetCapabilities.java

This is a follow up to [GEOS-8671](https://osgeo-org.atlassian.net/browse/GEOS-8671)

jometho:
Hi @tbarsballe is this fix for available in Geoserver 2.13.1?

tbarsballe:
Yes, it is; the associated PR was backported to GWC 1.13.x before the release.

PR #641, opened 2018-04-16
fix for AuthKey URLs 

ianturton:
This is a fix for https://github.com/GeoWebCache/geowebcache/issues/640 

It has no test but I'm not sure how to add a test without setting up an authkey server.


tbarsballe:
Was failing CI build, I restarted it and it seems fine. Merging.

tbarsballe:
Backported to 1.12.x and 1.13.x

PR #647, opened 2018-04-25
Remove GWC Home page requirement for XMLConfiguration

tbarsballe:
GWCDispatcher.appendStorageLocations() depends upon XMLConfiguration for legacy/compatibility reasons. This allows it to at least function when XMLConfiguration is unavailable.

Note: GeoServer will require a change in https://github.com/geoserver/geoserver/blob/master/src/gwc/src/main/resources/geowebcache-core-context.xml#L136, PR for that coming shortly.


tbarsballe:
GS PR here: https://github.com/geoserver/geoserver/pull/2855

PR #672, opened 2018-07-14
Prevent creating a Blobstore over top of other data. #671

smithkm:


geowebcache/core/src/test/java/org/geowebcache/util/FileMatchers.java
```diff
@@ -159,7 +163,7 @@ public void describeMismatch(Object item, Description description) {
             @Override
             public boolean matches(Object item) {
                 if(item instanceof File) {
-                    return ((File) item).isDirectory() && filesMatcher.matches(((File) item).listFiles());
+                    return ((File) item).isDirectory() && filesMatcher.matches(Arrays.asList(((File) item).listFiles()));
```
tbarsballe: Did this just not work before?
Seems like the sort of change that deserves its own unit test.

geowebcache/core/src/main/java/org/geowebcache/storage/UnsuitableStorageException.java
```diff
@@ -0,0 +1,45 @@
...
+ * @author smithkm
+ *
+ */
+public class UnsuitableStorageException extends StorageException {
+    
+    public UnsuitableStorageException(String msg) {
+        super(msg);
+    }
+
+    public static void checkSuitability(String location, final boolean exists, boolean empty)
```
tbarsballe: Its a bit weird for an exception to contain a utility method (even if that method does throw said exception), this seems like it would be more appropriate in CompositeBlobStore, same as where you define the thread local being tested

geowebcache/s3storage/src/test/java/org/geowebcache/s3/TemporaryS3Folder.java
```diff
@@ -129,7 +129,7 @@ public void delete() {
         checkState(isConfigured(), "client not configured.");
         if (temporaryPrefix == null) {
             return;
-        }
+        }
```
tbarsballe: Whitespace-only change

geowebcache/web/src/test/java/org/geowebcache/jetty/RestIntegrationTest.java
```diff
@@ -1042,6 +1047,104 @@ public void testLayerKillAll() throws Exception {
...
+                "<truncateLayer><layerName>"+layer2name+"</layerName></truncateLayer>";
+        
+        // Make it sure the blobstore doesn't exist already as that would invalidate the test
+        try (CloseableHttpResponse response = handleGet(URI.create("/geowebcache/rest/blobstores/maliciousCache.xml"), admin.getClient())){
+            assertThat(response, hasProperty("statusLine", hasProperty("statusCode", equalTo(404))));
+        }
+        
+        // Create a store
+        try (CloseableHttpResponse response = handlePut(URI.create("/geowebcache/rest/blobstores/maliciousCache"), admin.getClient(), blobStore)){
+            //assertThat(response, hasProperty("statusLine", hasProperty("statusCode", equalTo(201))));
```
tbarsballe: What's up with all these commented out asserts? 
I assume the exception is expected to be thrown from the handlePut, so shouldn't the body of the try have an Assert.fail()? Otherwise, uncomment or replace with a more descriptive comment.

geowebcache/core/src/test/java/org/geowebcache/util/FileMatchers.java
smithkm: No, it appears this was broken but wasn't being used.

geowebcache/web/src/test/java/org/geowebcache/jetty/RestIntegrationTest.java
smithkm: The response that comes back from these operations isn't important to the test, just that at the end the files that were there to start are still there. The commented out bits were to see what was going on while I set up the test and I should have deleted them.

geowebcache/core/src/main/java/org/geowebcache/storage/UnsuitableStorageException.java
tbarsballe: My comment here still stands

PR #674, opened 2018-07-23
#675 Support non-amazon s3 compatible servers in GWC S3 blobstore

NielsCharlier:


NielsCharlier:
Hey Kevin, I did run the existing tests with an endpoint configuration. A minio S3 server can be set up in barely two minutes (download the executable, run the executable, console gives you the credentials and endpoint) and then configured in the property file used by the tests. It was this that made me change the access rights test, because minio doesn't support it, and upgrade the aws-s3 lib version.

I will see what I can do for a mock S3 server,if it is not too time consuming I'm sure I can fit it in.

Where do I make a GWC ticket, GWC does not seem to be a project in the atlassian jira along with geotools,geoserver etc?

NielsCharlier:
I was able to make an offline test quite easily with the findify s3mock lib.

Unfortunately, this bug is affecting it:
https://github.com/findify/s3mock/issues/75

My suggestion would be to wait it out, since a patch is already under review. I leave the offline test in @Ignore for now in a number of months we can upgrade our dependency and enable the tests.

smithkm:
Thanks Niels, we use the GitHub ticket system for GWC. https://github.com/GeoWebCache/geowebcache/issues

NielsCharlier:
Here it is

PR #676, opened 2018-07-29
Changes to allow succesfull build on windows

aaime:
Changes to make the build pass on Windows. Themese:

- Use try-with-resources to make extra sure there are no leaks (a few critical bits were not covered)
- Make sure everything else needing to be released has a matching close/dispose/destroy
- Use nio Files utility class to perform create/delete operations in some bits to get a reason why files were not deleted or proper details on IO failure in general


aaime:
Great! @smithkm would you mind a AppVeyor build attached to pull requests like in GeoTools, to keep things building on Windows?

Issue #677, opened 2018-08-02
when using a minio endpoint empty prefix is not supported


NielsCharlier:
Some non-amazon s3 compatible servers, such as minio and s3mock struggle with leading slashes in the key. I previously thought this was only a problem with s3mock (and they do plan to resolve this), but minio also has a problem with it (which I didn't notice at first because the online test uses a prefix).

I am suggesting removing the leading slash from the key so that empty prefixes are more widely supported

jswk:
I've noticed that the leading slash is still present in path for the `_pending_deletes.properties` (in gwc-aws-s3 TMSKeyBuilder::pendingDeletes), which manifests itself when deleting a tile layer.

PR #678, opened 2018-08-02
when using a minio endpoint empty prefix is not supported #677

NielsCharlier:


geowebcache/s3storage/src/main/java/org/geowebcache/s3/TMSKeyBuilder.java
```diff
@@ -40,43 +41,43 @@
...
-            "%s/%s/" + PARAMETERS_METADATA_OBJECT_PREFIX;
+            "%s%s/" + PARAMETERS_METADATA_OBJECT_PREFIX;
 
     private String prefix;
 
     private TileLayerDispatcher layers;
 
     public TMSKeyBuilder(final String prefix, TileLayerDispatcher layers) {
-        this.prefix = prefix;
+        this.prefix = Strings.isNullOrEmpty(prefix) ? "" : prefix + "/";
```
smithkm: I really dislike how this confuses things.  Most of the `/`s are in the templates, but then this one special slash is being added here.  This is a really good opportunity for future mistakes.

Maybe this is a case for changing from using `String.format()` with templates that all do essentially the same thing for varying numbers of parameters to say just using `String.join()`.

NielsCharlier: @smithkm I didn't like it much either, but the only straight-forward alternatives I saw was to add an IF structure to every single method with either double the amount of templates or each time removing the first character. So I thought this was the least evil.

But I like your suggestion of String.join, I will change this PR.

NielsCharlier: The String.join method doesn't ignore nulls, in fact it will add the word "null" to it and still include an extra delimiter. Will have to write our own method.

smithkm:
```java
Arrays.stream(elements).filter(Objects::nonNull).collect(Collectors.joining("/"))
```

smithkm: Come to think of it, it should probably be escaping in case of elements containing slashes, but it's too late to add that as it would break backward compatibility for anyone who already has a layer or whatever with a slash in the name.

NielsCharlier:
@smithkm I changed it now. Thanks for having a look.

PR #683, opened 2018-08-15
[GEOS-8895] Fix Chunk[] is not a valid entry error for embedded GWC reload endpoint

tbarsballe:
https://osgeo-org.atlassian.net/browse/GEOS-8895

Applied the same fix as for GEOS-8474. Since this only affects embedded GeoWebCache, test case is in GeoServer: https://github.com/geoserver/geoserver/pull/3057


tbarsballe:
Could not reproduce the CI failures locally; rebased against master and re-pushed

PR #685, opened 2018-09-04
 Declare transitive spring dependencies in dependency-management

tbarsballe:
General improvements to the GWC root pom to better handle module dependencies (primarily transitive spring dependencies)

This hasn't yet caused an issue in core GWC, but should be cleaned up.


tbarsballe:
I'll backport to 1.13.x & 1.14.x

PR #689, opened 2018-09-27
GridSet method access to public for Geoserver Wfs3 vector tiles

fernandor777:
This PR modifies access level from protected to public for a method used in Geoserver WFS3 community module.

Geoserver PR:
https://github.com/geoserver/geoserver/pull/3117


aaime:
The method in question is exposed via GridSubset, which delegates it straight to GridSet. There is no good reason to have to build GridSubset to access this functionality, so exposing this method as public seems indeed an improvement. Merging.

PR #691, opened 2018-10-04
Updating to jai-ext 1.1.2 - backport to 1.14

dromagnoli:


aaime:
The RestIntegrationTest is unreliable and keeps on failing for whatever reason, @smithkm should we disable it? (does not even choose a random port, if 8080 is busy it will fail)

PR #694, opened 2018-10-10
Adding support for mapbox vector tiles new format, retaining ability to lookup the old one

aaime:


geowebcache/core/src/main/java/org/geowebcache/mime/ApplicationMime.java
```diff
@@ -59,10 +62,10 @@
 
     public static final ApplicationMime mapboxVector =
             new ApplicationMime(
-                    "application/x-protobuf",
+                    "application/vnd.mapbox-vector-tile",
                     "pbf",
-                    "mapbox-vectortile",
-                    "application/x-protobuf;type=mapbox-vector",
+                    "mapbox-vectortile-legacy",
```
tbarsballe: This changes the `internalName` of the *new *mapbox format to `mapbox-vector-legacy` - was this intentional?

This also breaks the demo page here: https://github.com/GeoWebCache/geowebcache/blob/master/geowebcache/core/src/main/java/org/geowebcache/demo/Demo.java#L390

PR #699, opened 2018-10-22
Remove com.sun.imageio.plugins.png.PNGImageWriter import, use static package name instead

tbarsballe:
Import was only used for an instanceof check, so static package name should be fine.

### DIFF
--- a/geowebcache/wms/src/main/java/org/geowebcache/io/ImageEncoderImpl.java
+++ b/geowebcache/wms/src/main/java/org/geowebcache/io/ImageEncoderImpl.java
@@ -14,7 +14,6 @@
  */
 package org.geowebcache.io;
 
-import com.sun.imageio.plugins.png.PNGImageWriter;
 import com.sun.media.imageioimpl.plugins.clib.CLibImageWriter;
 import it.geosolutions.imageio.stream.output.ImageOutputStreamAdapter;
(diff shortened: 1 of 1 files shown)


geowebcache/wms/src/main/java/org/geowebcache/io/ImageEncoderImpl.java
```diff
@@ -94,7 +93,9 @@ public ImageWriteParam prepareParameters(
                         // we can control quality here
                         params.setCompressionQuality(compressionRate);
                     }
-                } else if (writer instanceof PNGImageWriter) {
+                    // Use class name without import for JDK 11 compatibility
+                } else if ("com.sun.imageio.plugins.png.PNGImageWriter"
```
aaime: Technically one should go through all the superclasses and compare names (if it's a subclass of PNGImageWriter maybe?), but I guess this one is more than adequate.

tbarsballe: Technically true, although given its a sun internal class in the first place, with no official subclasses, I would hope no one is subclassing it.

aaime: Indeed

PR #702, opened 2018-10-24
Eliminate JDK11 XStream related warnings

aaime:
Same as in GeoServer, with unfortunate code duplication (there is no shared dependency where to put this), see https://github.com/geoserver/geoserver/pull/3198


aaime:
About having a module, yes, that would help, but thinking in terms of build dependendencies, geowebcache needs to be built before geoserver, and gs-main does depend only on geotools.
We could have a geowebcache module dedicated to xstream, or a geotools one.

My hope was to leave this duplication around and upgrade to a newer XStream version having less friction, and thus require less code duplication around.

PR #705, opened 2018-10-26
JDK 11 compatibility branch merge

aaime:


aaime:
Failure seems to be related to repo pollution with jars build by jdk11, marching on and merging

PR #711, opened 2018-11-01
Cleanup issues identified by SonarQube

smithkm:
Switch to using try-with-resources where opportunities have been identified by running static analysis with SonarQube.

This is all refactoring so there are no unit tests.


aaime:
Oooh, sonarqube... is it something available also for geotools and geoserver?

smithkm:
Boundless is using it for Boundless Server and its dependencies so yes it is being run on the other projects.  This was just the result of exploring how it works on my part.  I raised the idea of opening something to the community but the way we're currently using it that doesn't seem viable since it has a bunch of Boundless internal projects.

SonarQube rates the technical debts of the three projects at 1.5 kdays for GeoTools, 606 days for GeoServer, and 68 days for GWC.

geowebcache/diskquota/core/src/main/java/org/geowebcache/diskquota/UsageStatsMonitor.java
```diff
@@ -82,24 +82,32 @@ public void shutDown() {
...
+                                    + (attempts * seconds)
+                                    + " seconds. Attempt "
+                                    + attempts
+                                    + " of "
+                                    + maxAttempts
+                                    + "...";
+                    log.warn(message);
+                    if (attempts == maxAttempts) {
+                        throw new RuntimeException(message, e);
+                    }
```
tbarsballe: This whole block seems entirely identical to the one in QuotaUpdatesMonitor, above. Can it be moved to a shared superclass?
If not then perhaps update the error message to differentiate the two?

geowebcache/s3storage/src/main/java/org/geowebcache/s3/S3Ops.java
```diff
@@ -414,7 +414,7 @@ public Long call() throws Exception {
                         String.format(
                                 "S3 bulk delete aborted for '%s/%s'. Will resume on next startup.",
                                 bucketName, prefix));
-                return null;
+                throw e;
```
tbarsballe: Perhaps this is whats causing the AppVeyor test failure? Did you check how null results were handled from where this gets called?

geowebcache/diskquota/core/src/main/java/org/geowebcache/diskquota/UsageStatsMonitor.java
smithkm: Refactored to put most of the duplicate code in AbstractMonitor

geowebcache/s3storage/src/main/java/org/geowebcache/s3/S3Ops.java
smithkm: It seems to be one of the same errors occurring on other PRs

PR #712, opened 2018-11-06
Adding support for mapbox vector tiles new format, retaining abilityto lookup the old one (1.14.x)

tbarsballe:
1.14.x backport of https://github.com/GeoWebCache/geowebcache/pull/694


smithkm:
Have you tested doing an upgrade?  Does the old config still work, does it orphan existing cached tiles?

tbarsballe:
I have now tested an upgrade:

* Old config appears to get smoothly migrated to point to the new mime type (e.g. GWC demo page links to URL with the new format)
* If you use the new mime type, the old tile cache will get used; tiles are not orphaned.
* If you use the old mime type, the cache is ignored and new tiles are generated for each request
* The demo page no longer functions with either MVT mime type; looks like https://github.com/GeoWebCache/geowebcache/blob/master/geowebcache/core/src/main/java/org/geowebcache/demo/Demo.java#L390 was not updated to conform with these changed.

Note: All of the above, including the demo page issue, applies to master as well.

smithkm:
Well, for something we backport I would like to be able to ensure that client requests that would work before (including what mime type they request) would continue to hit existing seeded tiles after upgrading.  The client should not need to change, and the tiles should not need to be reseeded.

Ideally I'd want to avoid that even in a new version, but it would be OK with documentation.

tbarsballe:
Some updates:

I've got a fix for the various demo page issues here: https://github.com/GeoWebCache/geowebcache/pull/715 (master only, so far).

I'm having a bit of difficulty reproducing the cache miss issue:

> * If you use the old mime type, the cache is ignored and new tiles are generated for each request

That may have just been caused by some interference / confusion introduced by various issues with the demo page.

tbarsballe:
I've now tested the new oversampled VT code with an old (not oversampled) tile cache, and the old tiles are preserved, so that part at least is backwards compatible.

tbarsballe:
Updated with demo-page fixes

PR #717, opened 2018-11-19
Restore and update the maven jetty plugin configuration

tbarsballe:
This was incorrectly removed during the Spring MVC REST update.
I've restored it, and updated it to a modern version of the plugin.

As before, it can be run using `mvn jetty:run`


tbarsballe:
AWS S3 is failing on this PR - I think there might be an issue with the current master, since this shouldn't affect it at all

PR #723, opened 2019-01-08
Fix: Support utf8 for html

ATLgo:
support utf8 because of chinese can't be used.


aaime:
Build failed due to formatting error, can be fixed by running "mvn fmt:format" on the sources and committing the result. Can you do it?

ATLgo:
> Build failed due to formatting error, can be fixed by running "mvn fmt:format" on the sources and committing the result. Can you do it?
I Fixed it again with new PR #854

PR #732, opened 2019-02-02
Spotbugs integration

aaime:
Follows up (but does not depend on) https://github.com/geotools/geotools/pull/2239


aaime:
Awesome, one in then

PR #747, opened 2019-04-14
Deprecation cleanup

aaime:
Makes deprecated call fail the build in QA mode, removes or explicitly ignores deprecated calls, removes some of the deprecated methods/classes. 

Follows up https://github.com/geotools/geotools/pull/2309


aaime:
Please do not merge yet, some of the deprecations have been made carelessly and GeoServer depends on them....

geowebcache/core/src/main/java/org/geowebcache/util/IOUtils.java
```diff
@@ -0,0 +1,35 @@
...
+import java.io.IOException;
+
+/** IO related utility methods that common libraries won't provide */
+public class IOUtils {
+
+    /**
+     * A replacement for commons-io closeQuietly, for those rare cases in which the quiet closing
+     * behavior is actually needed and try-with-resources won't do the expected job
+     */
+    public static void closeQuietly(Closeable clo) {
```
jodygarnett: Is there not a GeoTools method for this?

geowebcache/core/src/test/java/org/geowebcache/TestHelpers.java
```diff
@@ -30,7 +31,8 @@
  */
 public class TestHelpers {
 
-    static GridSetBroker gridSetBroker = new GridSetBroker(false, false);
+    static GridSetBroker gridSetBroker =
+            new GridSetBroker(Collections.singletonList(new DefaultGridsets(false, false)));
```
jodygarnett: This is such a common snippet was it inlined? It may be worth having a new GridSetBroker() constructor that is configured with a singleton list as illustrated above.

geowebcache/core/src/main/java/org/geowebcache/util/IOUtils.java
aaime: GeoWebCache does not depend on GeoTools all that much, might be that the right jar was not on the path. That said, I grepped geootols for a closeQuietly and did not find one. GeoServer has one though.

geowebcache/core/src/test/java/org/geowebcache/TestHelpers.java
aaime: The constructor was deprecated, with instructions to do exactly the above:
https://github.com/GeoWebCache/geowebcache/blob/112060ad43708c030085fbbb147e6437a5a7e558/geowebcache/core/src/main/java/org/geowebcache/grid/GridSetBroker.java#L56

If each dev coming in tries to apply his on personal preference we'd just get chaos, I've tried not to question too much deprecations, unless they were evidently broken, .
That said, to be honest, I did not like this one, and a few others in GWC, that made for pretty tangled up replacement code especially in the GeoServer.

PR #751, opened 2019-05-06
Updating to imageio-ext-1.1.29

dromagnoli:


dromagnoli:
Failure is due to gwc-diskquota -jdbc not proceeding after 10 minutes.
It should be unrelated with the pom update.
So I will merge it anyway

PR #754, opened 2019-05-10
prevent focus steal on macOS

jodygarnett:
Just setting the property at the top of the file does not mean it is passed down to the surefire and failsafe plugins.


tbarsballe:
Looks good to me. AppVeyor id failing for what seems like no reason.

PR #756, opened 2019-05-13
issue#755-support TMS FlipY

dromagnoli:


geowebcache/tms/src/main/java/org/geowebcache/service/tms/TMSService.java
```diff
@@ -178,10 +178,29 @@ public ConveyorTile getConveyor(HttpServletRequest request, HttpServletResponse
         }
 
         String[] yExt = params[paramsLength - 1].split("\\.");
+        String x = params[paramsLength - 2];
+        String y = yExt[0];
+        String z = params[paramsLength - 3];
+
+        // look for flipY parameter and do flipping if flipY=true
+        String flip = yExt[1].toUpperCase();
+        if (flip.contains("?FLIPY=TRUE")) {
```
aaime: This seems a fragile way to handle query parameters. What if for example one would make a request like 
"...?cacheBreaker=randomeNumber&flipy=true"?
Instead, I would throw away anything that is beyond the ? and use ``getParameterNames`` and ``getParameter`` to grab the eventual flipy parameter.

geowebcache/tms/src/main/java/org/geowebcache/service/tms/TMSService.java
```diff
@@ -178,10 +178,29 @@ public ConveyorTile getConveyor(HttpServletRequest request, HttpServletResponse
...
+            String[] yFlip = yExt[1].split("\\?");
+            int zInt = Integer.parseInt(z);
+            int flippedY = Integer.parseInt(y);
+
+            // recompute the yValue doing the flipping:
+            // flippedY = 2^level - y - 1
+            flippedY = (int) (Math.pow(2, zInt)) - flippedY - 1;
```
aaime: This embeds the assumption that the first level has a single tile... which is normally the case, but I'm not sure, it could be possible for one to define a base level with 2 tiles stacked vertically for a CRS that has an elongated area (similar to how WGS84 has two horizontal tiles at the first zoom level). The WMTS machinery seems to have a more general approach to flip the Y value: https://github.com/GeoWebCache/geowebcache/blob/760b72177dd99dacbd71ccbf7862ec9f6c5e3304/geowebcache/wmts/src/main/java/org/geowebcache/service/wmts/WMTSService.java#L465

dromagnoli:
@aaime I have updated the PR with the new logic to address your feedbacks.
Please, let me know.

dromagnoli:
The appveyor is failing on the usual gwc-aws-s3 test failure.
Going to merge

PR #763, opened 2019-05-21
issue#755-support TMS FlipY - backport to 1.15.x

dromagnoli:
The change has no effects when not using the specific flag so the backport should be safe.


dromagnoli:
Now it's an MBTiles test failing (not related to this change)

PR #776, opened 2019-07-08
Azure blobstore backport to 1.15.x

aaime:
The backport per se is trivial, but doing it I noticed that all the blobstores are actually part of the final WAR file, while I did not add the azure one to it.
Thinking out loud, should I:

- Add it to the GWC war on master, since it's a dev series
- Skip adding in on 1.15.x (or make it available as a profile), since it's a stable series 

@smithkm what do you think?


aaime:
Moving on with the merge in the meantime, as a module not included in the final war it's not going to do harm :-D

Issue #777, opened 2019-07-16
Allow runtime parametrization of Azure blobstore configuration


aaime:
Right now the eventual enviroment parametrization of the configuration (e.g., usage of ${param} in the config) is resolved at the XStream loading level. 
This works fine for the hand editing of files, but breaks both the REST API and the GeoServer UI editing:

- In the case of REST API, the BlobStoreInfo is read, and the env variables resolved during reading, meaning that saving it would save the resolved information, not the variables
- In the case of the GeoServer UI see https://osgeo-org.atlassian.net/browse/GEOS-9288 

The simplest approach is to modify the  BlobStoreInfo so that it contains only attributes of type string, with the same names as today, and remove the env variable resolution from the XStream configuration.

The variables will then be resolved when ``BlobStoreInfo.createBlobStore`` is called.

The downsides of this approach would be:

- Less explicit typing, the type checks need to be performed during env variable resolution
- Would change the interface of the AzureBlobStoreInfo, but it's new, so no particular problem here

The approach can then eventually be extended to other BlobStoreInfo, as the main configuration interfaces (direct XML editing, REST and GeoServer GUI) would not be broken by the change, only direct programmatic usage would be (which is actually not present in our current code, and would seem unlikely even outside of it given the presence of a REST API for configuration).

aaime:
Solved by #779

PR #779, opened 2019-07-17
[#777] Allow runtime parametrization of Azure blobstore configuration

fernandor777:
This PR fixes runtime parametrization on Azure blobstore configuration, leveraging placeholders resolving on blobstore creation runtime instead of Xstream resolvers.

Issue:
https://github.com/GeoWebCache/geowebcache/issues/777

### DIFF
--- a/geowebcache/azureblob/src/main/java/org/geowebcache/azure/AzureBlobStoreConfigProvider.java
+++ b/geowebcache/azureblob/src/main/java/org/geowebcache/azure/AzureBlobStoreConfigProvider.java
@@ -14,90 +14,16 @@
  */
 package org.geowebcache.azure;
 
-import com.google.common.base.Strings;
 import com.thoughtworks.xstream.XStream;
-import com.thoughtworks.xstream.converters.SingleValueConverter;
-import com.thoughtworks.xstream.converters.basic.BooleanConverter;
-import com.thoughtworks.xstream.converters.basic.IntConverter;
-import com.thoughtworks.xstream.converters.basic.StringConverter;
-import org.geowebcache.GeoWebCacheEnvironment;
-import org.geowebcache.GeoWebCacheExtensions;
-import org.geowebcache.config.BlobStoreInfo;
 import org.geowebcache.config.Info;
 import org.geowebcache.config.XMLConfigurationProvider;
 
 public class AzureBlobStoreConfigProvider implements XMLConfigurationProvider {
 
-    private static GeoWebCacheEnvironment gwcEnvironment = null;
-
-    private static SingleValueConverter EnvironmentNullableIntConverter =
-            new IntConverter() {
-
-                @Override
-                public Object fromString(String str) {
-                    str = resolveFromEnv(str);
-                    if (Strings.isNullOrEmpty(str)) {
-                        return null;
-                    }
-                    return super.fromString(str);
-                }
-            };
-
-    private static SingleValueConverter EnvironmentNullableBooleanConverter =
-            new BooleanConverter() {
-
-                @Override
-                public Object fromString(String str) {
-                    str = resolveFromEnv(str);
-                    if (Strings.isNullOrEmpty(str)) {
-                        return null;
-                    }
-                    return super.fromString(str);
-                }
-            };
-
-    private static SingleValueConverter EnvironmentStringConverter =
-            new StringConverter() {
-                @Override
-                public Object fromString(String str) {
-                    str = resolveFromEnv(str);
-                    if (Strings.isNullOrEmpty(str)) {
-                        return null;
-                    }
-                    return str;
-                }
-            };
-
-    private static String resolveFromEnv(String str) {
-        if (gwcEnvironment == null) {
-            gwcEnvironment = GeoWebCacheExtensions.bean(GeoWebCacheEnvironment.class);
-        }
-        if (gwcEnvironment != null
-                && str != null
-                && GeoWebCacheEnvironment.ALLOW_ENV_PARAMETRIZATION) {
-            Object result = gwcEnvironment.resolveValue(str);
-            if (result == null) {
-                return null;
-            }
-            return result.toString();
-        }
-        return str;
-    }
-
     @Override
     public XStream getConfiguredXStream(XStream xs) {
         Class<AzureBlobStoreInfo> clazz = AzureBlobStoreInfo.class;
         xs.alias("AzureBlobStore", clazz);
-        xs.registerLocalConverter(clazz, "maxConnections", EnvironmentNullableIntConverter);
-        xs.registerLocalConverter(clazz, "proxyPort", EnvironmentNullableIntConverter);
-        xs.registerLocalConverter(clazz, "useHTTPS", EnvironmentNullableBooleanConverter);
-        xs.registerLocalConverter(clazz, "container", EnvironmentStringConverter);
-        xs.registerLocalConverter(clazz, "accountName", EnvironmentStringConverter);
-        xs.registerLocalConverter(clazz, "accountKey", EnvironmentStringConverter);
-        xs.registerLocalConverter(clazz, "prefix", EnvironmentStringConverter);
-        xs.registerLocalConverter(clazz, "proxyHost", EnvironmentStringConverter);
-        xs.registerLocalConverter(
-                BlobStoreInfo.class, "enabled", EnvironmentNullableBooleanConverter);
         xs.aliasField("id", clazz, "name");
         return xs;
     }
--- a/geowebcache/azureblob/src/main/java/org/geowebcache/azure/AzureBlobStoreData.java
+++ b/geowebcache/azureblob/src/main/java/org/geowebcache/azure/AzureBlobStoreData.java
@@ -1,0 +1,182 @@
+/**
+ * This program is free software: you can redistribute it and/or modify it under the terms of the
+ * GNU Lesser General Public License as published by the Free Software Foundation, either version 3
+ * of the License, or (at your option) any later version.
+ *
+ * <p>This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * <p>You should have received a copy of the GNU Lesser General Public License along with this
+ * program. If not, see <http://www.gnu.org/licenses/>.
+ *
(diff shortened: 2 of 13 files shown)


aaime:
QA checks failed:

```
[WARNING] /home/travis/build/GeoWebCache/geowebcache/geowebcache/core/src/main/java/org/geowebcache/GeoWebCacheTypedEnvironment.java:[49,36] Integer(java.lang.String) in java.lang.Integer has been deprecated
[WARNING] /home/travis/build/GeoWebCache/geowebcache/geowebcache/core/src/main/java/org/geowebcache/GeoWebCacheTypedEnvironment.java:[59,33] Boolean(java.lang.String) in java.lang.Boolean has been deprecated
```

In both cases use the valueOf static method instead, these constructors are bad for performance and are about to be removed in newer versions of Java.

To run these checks locally, add "-Dqa" to the maven call (warning, slow, best do it only in the module that got modified). This is valid for gwc but also gt and gs.

geowebcache/azureblob/src/main/java/org/geowebcache/azure/AzureBlobStoreConfigProvider.java
```diff
@@ -60,36 +41,19 @@ public Object fromString(String str) {
...
-        return str;
-    }
-
     @Override
     public XStream getConfiguredXStream(XStream xs) {
         Class<AzureBlobStoreInfo> clazz = AzureBlobStoreInfo.class;
         xs.alias("AzureBlobStore", clazz);
-        xs.registerLocalConverter(clazz, "maxConnections", EnvironmentNullableIntConverter);
-        xs.registerLocalConverter(clazz, "proxyPort", EnvironmentNullableIntConverter);
+        xs.registerLocalConverter(clazz, "maxConnections", EnvironmentStringConverter);
```
aaime: No need to register custom converters any longer, the default behavior should do (please check).

geowebcache/azureblob/src/main/java/org/geowebcache/azure/AzureBlobStoreConfigProvider.java
```diff
@@ -18,37 +18,18 @@
...
-                public Object fromString(String str) {
-                    str = resolveFromEnv(str);
-                    if (Strings.isNullOrEmpty(str)) {
-                        return null;
-                    }
-                    return super.fromString(str);
-                }
-            };
-
     private static SingleValueConverter EnvironmentNullableBooleanConverter =
```
aaime: Likely not needed any longer, just check the default behavior in the case of "empty or null"

geowebcache/azureblob/src/main/java/org/geowebcache/azure/AzureBlobStoreInfo.java
```diff
@@ -41,7 +41,7 @@
      */
     public static final int DEFAULT_CONNECTIONS = 100;
 
-    static Log log = LogFactory.getLog(AzureBlobStoreInfo.class);
+    private static final Logger LOGGER = Logger.getLogger(AzureBlobStoreInfo.class.getName());
```
aaime: GWC uniformly uses commons logging, please revert this change.

geowebcache/azureblob/src/main/java/org/geowebcache/azure/AzureBlobStoreInfo.java
```diff
@@ -123,12 +123,14 @@ public void setPrefix(String prefix) {
     }
 
     /** @return The maximum number of allowed open HTTP connections. */
-    public Integer getMaxConnections() {
-        return maxConnections == null ? DEFAULT_CONNECTIONS : maxConnections;
+    public String getMaxConnections() {
+        return maxConnections == null
+                ? Integer.valueOf(DEFAULT_CONNECTIONS).toString()
```
aaime: ``String.valueOf(DEFAULT_CONNECTIONS)`` would be more straightforward.

geowebcache/azureblob/src/main/java/org/geowebcache/azure/AzureBlobStoreData.java
```diff
@@ -0,0 +1,190 @@
...
+
+import java.net.InetSocketAddress;
+import java.net.Proxy;
+import org.geowebcache.GeoWebCacheTypedEnvironment;
+
+/**
+ * Azure Blobstore type-resolved data from a {@link AzureBlobStoreInfo} using enviroment variables
+ * if enabled.
+ */
+public class AzureBlobStoreData {
```
aaime: Type safe bean, yep, I would have likely done the same. Any actual need to make this class public, though?

geowebcache/azureblob/src/main/java/org/geowebcache/azure/AzureBlobStoreData.java
```diff
@@ -0,0 +1,190 @@
...
+
+    public Integer getMaxConnections() {
+        return maxConnections;
+    }
+
+    public void setMaxConnections(Integer maxConnections) {
+        this.maxConnections = maxConnections;
+    }
+
+    public Boolean isUseHTTPS() {
```
aaime: Should allow parametrization of this one too --> String.

geowebcache/azureblob/src/test/java/org/geowebcache/azure/AzureBlobStoreConfigProviderTest.java
```diff
@@ -18,35 +18,10 @@
...
-import org.geowebcache.GeoWebCacheExtensions;
-import org.junit.After;
-import org.junit.Before;
 import org.junit.Test;
-import org.springframework.context.support.ClassPathXmlApplicationContext;
 
 public class AzureBlobStoreConfigProviderTest {
 
-    @Before
-    public void setUp() throws Exception {
```
aaime: Please do not remove, the test is now to check that the environment variables can survive a xstream load (this is actually at the very core of the issue report), while there should be another test checking that variable expansion works.

geowebcache/azureblob/src/test/java/org/geowebcache/azure/AzureBlobStoreDataTest.java
```diff
@@ -0,0 +1,65 @@
...
+        GeoWebCacheExtensions gse = new GeoWebCacheExtensions();
+        gse.setApplicationContext(context);
+    }
+
+    @After
+    public void tearDown() throws Exception {
+        System.clearProperty("CONTAINER");
+        System.clearProperty("CONNECTIONS");
+        System.clearProperty("ACCOUNT_NAME");
+        System.clearProperty("ALLOW_ENV_PARAMETRIZATION");
```
aaime: Should clear ACCOUNT_KEY too

geowebcache/core/src/main/java/org/geowebcache/GeoWebCacheTypedEnvironment.java
```diff
@@ -0,0 +1,69 @@
...
+ *
+ * @author Fernando Mino, GeoSolutions S.A.S., Copyright 2019
+ */
+package org.geowebcache;
+
+import java.util.Optional;
+import org.apache.commons.lang3.StringUtils;
+
+/** Enviroment variables placeholder resolver with declared types. */
+public class GeoWebCacheTypedEnvironment {
```
aaime: I don't see a need for a new class, an extra static method in GeoWebCacheEnvironment should do the trick.

geowebcache/core/src/main/java/org/geowebcache/GeoWebCacheTypedEnvironment.java
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    /**
+     * Resolve placeholders for a String parameter if enabled on configurations.
+     *
+     * @param value String value with optional placeholders.
+     * @param type Data Type to return
+     * @return Optional resolved value.
+     */
+    public <T extends Object> Optional<T> resolve(final String value, Class<T> type) {
```
aaime: I'd keep method name conventions intact, call it ``resolveValue`` like in ``GeoWebCacheEnvironment`` and it has the extra parameter for type conversion.

geowebcache/azureblob/src/main/java/org/geowebcache/azure/AzureBlobStoreConfigProvider.java
fernandor777: Indeed not needed anymore, removed.

geowebcache/azureblob/src/main/java/org/geowebcache/azure/AzureBlobStoreInfo.java
fernandor777: Thanks, reverted.

geowebcache/azureblob/src/main/java/org/geowebcache/azure/AzureBlobStoreInfo.java
fernandor777: Fixed.

geowebcache/azureblob/src/main/java/org/geowebcache/azure/AzureBlobStoreData.java
fernandor777: Right, not public makes sense.

geowebcache/azureblob/src/test/java/org/geowebcache/azure/AzureBlobStoreDataTest.java
fernandor777: Added.

geowebcache/core/src/main/java/org/geowebcache/GeoWebCacheTypedEnvironment.java
fernandor777: Indeed, merged new method on GeoWebCacheEnvironment class.

geowebcache/core/src/main/java/org/geowebcache/GeoWebCacheTypedEnvironment.java
fernandor777: Added as resolveValueIfEnabled since this method does the check for parameter-replacement flag enabled inside (the former 'resolveValue' method doesn't).

geowebcache/azureblob/src/test/java/org/geowebcache/azure/AzureBlobStoreConfigProviderTest.java
fernandor777: Reverted.

aaime:
Looks good, merged and hopefully kicked an automatic backport.

PR #781, opened 2019-07-19
Add text/mapml as an infoformat for use with GeoServer MapML module

prushforth:
The text/mapml info format is hidden if the MapML Community Module is not installed.


prushforth:
Thanks!

PR #791, opened 2019-09-24
Improving truncate performance on tile ranges 

rigolepe:
We have geowebcache running in a production environment and we notice very slow truncate operations on our tiles in a regular file blob store. The culprit is the `storageBroker.delete(tileRange);`  call in the `TruncateTask`, because the underlying implementation seems to iterate through all the tiles and checks for each one of them if they are located in the given tile range. See the frequent use of `listFilesNullSafe` in `FileBlobStore`: 

```
public class FileBlobStore implements BlobStore {
...
public boolean delete(TileRange trObj) throws StorageException {
...
final FilePathFilter tileFinder = new FilePathFilter(trObj);
...
File[] srsZoomDirs = listFilesNullSafe(layerPath, tileFinder);
...
File[] intermediates = listFilesNullSafe(srsZoomParamId, tileFinder);
...
File[] tiles = listFilesNullSafe(imd, tileFinder);
...
```

We cache our tiles up to zoom level 10 in a 240km x 100km area. Truncating even a very small area results in a noticeable performance hit on our geowebcache with a truncate task running minutes on end.

If we change the implementation of `TruncateTask` to reflect the implementation of the `SeedTask` regarding tile location so that we eventually only delete the actual storage objects that are located in the given truncate range: 
```
storageBroker.delete(tile.getStorageObject());
```
instead of 
```
storageBroker.delete(tileRange);
```
After this change, we see a remarkable improvement in our truncate times (milliseconds instead of minutes). See our pull request for the details. We tested this implementation on our file store, but not on other storage types, but all tests succeed anyway. 

In order to analyse this problem, we also added several `toString()` methods and we left them in te pull request. We can remove them if that is desired. 

The implementation details that are now shared between `SeedTask` and `TruncateTask` can probably be refactored away in a common method or so, but we didn't want to complicate the pull request for now.


geowebcache/core/src/main/java/org/geowebcache/seed/TruncateTask.java
```diff
@@ -49,11 +56,40 @@ public TruncateTask(StorageBroker sb, TileRange tr, TileLayer tl, boolean doFilt
...
-            super.state = GWCTask.STATE.DEAD;
-            log.error("During truncate request", e);
+
+        final String layerName = tl.getName();
+        TileRange tr = trIter.getTileRange();
+        checkInterrupted();
+
+        long[] gridLoc = trIter.nextMetaGridLocation(new long[3]);
+
+        while (gridLoc != null && !this.terminate) {
```
aaime: Won't this make the code iterate over tiles that do not exist? What if one truncates a EPSG:900913 with 30 zoom levels,
won't this take ages to finish? The current code won't even visit a zoom level directory if it's not there.

geowebcache/core/src/main/java/org/geowebcache/seed/TruncateTask.java
```diff
@@ -49,11 +56,40 @@ public TruncateTask(StorageBroker sb, TileRange tr, TileLayer tl, boolean doFilt
...
+                checkInterrupted();
+                storageBroker.delete(tile.getStorageObject());
+            } catch (Exception e) {
+                super.state = GWCTask.STATE.DEAD;
+                log.error("During truncate request", e);
+                throw new GeoWebCacheException(e);
+            }
+
+            checkInterrupted();
+            gridLoc = trIter.nextMetaGridLocation(gridLoc);
```
aaime: Going over meta grid locations means deleting only one tile for each metatile, leaving all others intact... or not?

aaime:
Did a quick check by putting a breakpoint in the FileBlobStore.deleteRange code, it ends up reading only the existing dirs... I believe I see what the difference is. The current code favours big truncates, by only scanning what's actually on the file system, but ends up listing also tiles that are not in the range in the process. The TileFinder does a good job at filtering out the tiles that are not in the requested zoom levels, but it's not filtering on the intermediate path, thus listing all possible intermediate directories and ending up listing each single tile... that's really bad for small truncates.

However, the changes you proposed are not acceptable:
- would kill performance large truncates
- does not seem to delete all tiles that needs to be removed (only one every metatile)

Suggestion: start over and implement proper intermediate directory filtering in TileFinder instead.

aaime:
Hey I was working on supporting XYZ tile layouts on file system and gave a crack at better filtering of tiles on the default layout as well. Want to give it a look and maybe a try?
https://github.com/GeoWebCache/geowebcache/pull/795

rigolepe:
Hi, thanks for the feedback! 
I will have a closer look at this in the coming weeks.

PR #795, opened 2019-10-27
Support TMS and XYZ tile layouts on file system

aaime:
Allows using the file blob store with the TMS and XYZ tiles/directory layouts, for static seeding.
Also optimizes directory lookup on default file layout for faster tile truncations.

### DIFF
--- a/geowebcache/core/src/main/java/org/geowebcache/config/FileBlobStoreInfo.java
+++ b/geowebcache/core/src/main/java/org/geowebcache/config/FileBlobStoreInfo.java
@@ -21,7 +21,9 @@ import org.geowebcache.locks.LockProvider;
 import org.geowebcache.storage.BlobStore;
 import org.geowebcache.storage.BlobStoreListener;
 import org.geowebcache.storage.StorageException;
+import org.geowebcache.storage.blobstore.file.DefaultFilePathGenerator;
 import org.geowebcache.storage.blobstore.file.FileBlobStore;
+import org.geowebcache.storage.blobstore.file.XYZFilePathGenerator;
 
 /**
  * Configuration and factory for {@link FileBlobStore}.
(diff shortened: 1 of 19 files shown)


geowebcache/core/src/main/java/org/geowebcache/storage/blobstore/file/DefaultFilePathGenerator.java
```diff
@@ -0,0 +1,133 @@
...
+ * of the License, or (at your option) any later version.
+ *
+ * <p>This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * <p>You should have received a copy of the GNU Lesser General Public License along with this
+ * program. If not, see <http://www.gnu.org/licenses/>.
+ *
+ * @author Arne Kepp, The Open Planning Project, Copyright 2008
```
jodygarnett: new file?

geowebcache/core/src/test/java/org/geowebcache/blobstore/file/DefaultFilePathGeneratorTest.java
```diff
@@ -0,0 +1,146 @@
+package org.geowebcache.blobstore.file;
```
jodygarnett: header

geowebcache/core/src/main/java/org/geowebcache/config/FileBlobStoreInfo.java
```diff
@@ -31,12 +33,20 @@
  */
 public class FileBlobStoreInfo extends BlobStoreInfo {
 
+    public static enum PathGeneratorType {
+        DEFAULT,
+        TMS,
+        SLIPPY
```
jodygarnett: I would go with XYZ just to be a bit more self documenting

geowebcache/core/src/main/java/org/geowebcache/storage/blobstore/file/DefaultFilePathGenerator.java
aaime: It's code that Arne wrote regardess, just moved around.

geowebcache/core/src/main/java/org/geowebcache/config/FileBlobStoreInfo.java
aaime: Renamed

geowebcache/core/src/test/java/org/geowebcache/blobstore/file/DefaultFilePathGeneratorTest.java
aaime: Added

PR #801, opened 2019-11-17
Removing maven build warnings

aaime:
The current maven build complains about lack of versions in some plugins, duplicate dependency definitions, and lack of file encoding indication in reporting. Cleaning those away.

### DIFF
--- a/geowebcache/core/pom.xml
+++ b/geowebcache/core/pom.xml
@@ -176,11 +176,6 @@
       <scope>test</scope>
     </dependency>
     <dependency>
-      <groupId>org.springframework</groupId>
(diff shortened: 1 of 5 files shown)


jodygarnett:
Andrea this has a build failure for travis: PMD is failing to run due to a missing class.

aaime:
Ah, it happens because of corrupted jars in the local repo... will need to clean it up...

jodygarnett:
I am ... confused by your statement. Corrupted jars on your machine, or on that travis node?

aaime:
Had to go into travis and remove all build caches (tried getting rid of specific ones with no luck), that allowed fresh jars to be downloaded an problem solved.

PR #802, opened 2019-11-20
Using trusty to obtain oraclejdk8 environment

jodygarnett:


jodygarnett:
The travis build above fails ... on `AWS S3 Storage` <-- wot?

jodygarnett:
This shows a solid consistent oraclejdk8 failure on S3 support

aaime:
Some of the S3 tests are, simply put, unreliable, using MT behavior and just adding some waits to compensate for the uncertainty. In some environments they fail more often than others. Patches welcomed.

aaime:
@jodygarnett so what do we do about this one? Do you want to merge or not?

jodygarnett:
This would switch to openjdk right? I am fine with that ...

aaime:
I don't know that it would no, but I could be wrong.

aaime:
Ok I have a PR that's failing only on 1.15.x and indeed it has a Oracle JDK issue.
However, master and 1.16.x, working, have a different setup:

```
sudo: false
cache:
  directories:
    - "$HOME/.m2"
language: java
before_script:
  - cd geowebcache
script:
  - mvn -B -U -T2 -fae clean install -Dtest.maxHeapSize=512M -Djava.awt.headless=true $ARGS
notifications:
  email:
    on_success: never
    on_failure: never
matrix:
  include:
    - jdk: openjdk8
      env: ARGS="-Dfmt.skip=true"
    - jdk: openjdk11
      env: ARGS="-Dfmt.skip=true"
    - jdk: openjdk11
      env: ARGS="-Dfmt.action=check -Dqa -DskipTests=true"
```

So asking not for a particular distro, but explicitly for openjdk8. I think this PR should be closed and replaced by one that aligns 1.15.x with the other two branches. @jodygarnett ?

jodygarnett:
Closing this experiment

PR #806, opened 2019-12-03
Mass Truncate for GWC Cache

imranrajjad:
-added support to clear entire GWC cache through REST and GUI
-updated documentation

mass truncate can be invoked from REST using

`curl -v -u geowebcache:secured -X POST -H "Content-type: text/xml" -d "<truncateAll></truncateAll>"  "http://localhost:8080/geowebcache/rest/masstruncate"`

Demo page GUI also includes a link to invoke Mass Truncate of all layers

Issue discussed at : https://github.com/geosolutions-it/support/issues/381

### DIFF
--- a/geowebcache/core/src/main/java/org/geowebcache/seed/TruncateAllRequest.java
+++ b/geowebcache/core/src/main/java/org/geowebcache/seed/TruncateAllRequest.java
@@ -24,6 +24,9 @@ import org.geowebcache.GeoWebCacheException;
 import org.geowebcache.layer.TileLayer;
 import org.geowebcache.storage.StorageBroker;
 import org.geowebcache.storage.StorageException;
+import org.geowebcache.util.ServletUtils;
+import org.springframework.http.HttpStatus;
+import org.springframework.http.ResponseEntity;
 
 /** @author ImranR */
 @XStreamAlias("truncateAll")
@@ -41,26 +44,34 @@ public class TruncateAllRequest implements MassTruncateRequest, Serializable {
             throws StorageException, GeoWebCacheException {
         Iterator<TileLayer> iterator = breeder.getLayers().iterator();
         TileLayer toTruncate;
+        Iterator<String> gridSetIterator;
+        String gridSetId = "";
         boolean truncated = false;
         while (iterator.hasNext()) {
             toTruncate = iterator.next();
-            truncated = sb.delete(toTruncate.getName());
-            if (!truncated) {
-                // did we hit a layer that has nothing on storage, or a layer that is not there?
-                try {
-                    breeder.findTileLayer(toTruncate.getName());
-                } catch (GeoWebCacheException e) {
-                    throw new IllegalArgumentException(
-                            "Could not find layer " + toTruncate.getName());
-                }
-            } else {
-                log.info("Truncate All Job: Finished deleting layer " + toTruncate.getName());
-                if (getTrucatedLayers().length() > 0) getTrucatedLayers().append(",");
+            // get all grid sets
+            gridSetIterator = toTruncate.getGridSubsets().iterator();
+            while (gridSetIterator.hasNext()) {
+                gridSetId = gridSetIterator.next();
+                truncated = sb.deleteByGridSetId(toTruncate.getName(), gridSetId);
+                log.info("Layer: " + toTruncate.getName() + ",Truncated Gridset :" + gridSetId);
+            }
+            if (truncated) {
+                if (getTrucatedLayers().length() > 0) trucatedLayers.append(",");
                 getTrucatedLayers().append(toTruncate.getName());
             }
         }
 
         return true;
+    }
+
+    @Override
+    public ResponseEntity<String> getResponse(String contentType) {
+        // for gui send page
+        // for others stay legacy
+        if (contentType.equalsIgnoreCase("application/x-www-form-urlencoded")) {
+            return new ResponseEntity<String>(getResponsePage().toString(), HttpStatus.OK);
+        } else return MassTruncateRequest.super.getResponse(contentType);
     }
 
     public StringBuilder getTrucatedLayers() {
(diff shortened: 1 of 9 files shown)


imranrajjad:
after merging this, please re-kick the build on https://github.com/geoserver/geoserver/pull/3922

geowebcache/core/src/main/java/org/geowebcache/seed/TruncateAllRequest.java
```diff
@@ -0,0 +1,76 @@
...
+        boolean truncated = false;
+        while (iterator.hasNext()) {
+            toTruncate = iterator.next();
+            truncated = sb.delete(toTruncate.getName());
+            if (!truncated) {
+                // did we hit a layer that has nothing on storage, or a layer that is not there?
+                try {
+                    breeder.findTileLayer(toTruncate.getName());
+                } catch (GeoWebCacheException e) {
+                    throw new IllegalArgumentException(
```
aaime: The user did not specify any layer name, just asked for a "clean all", no reason to throw an exception, log a warning and move on instead.

geowebcache/core/src/main/java/org/geowebcache/seed/TruncateAllRequest.java
```diff
@@ -0,0 +1,76 @@
...
+
+    @Override
+    public boolean doTruncate(StorageBroker sb, TileBreeder breeder)
+            throws StorageException, GeoWebCacheException {
+        Iterator<TileLayer> iterator = breeder.getLayers().iterator();
+        TileLayer toTruncate;
+        boolean truncated = false;
+        while (iterator.hasNext()) {
+            toTruncate = iterator.next();
+            truncated = sb.delete(toTruncate.getName());
```
aaime: One more thing, does disk quota on the layer in question work fine after this call? I'm asking because GeoServer does the layer truncation by iterating on its gridset ids, but I see the TruncateLayer class here also does it this way.
Wondering if listeners might be thrown off thinking the layer is gone, rather than its tiles, have a quick look please.

geowebcache/rest/src/main/java/org/geowebcache/rest/controller/MassTruncateController.java
```diff
@@ -146,6 +148,45 @@ public void setTileBreeder(TileBreeder breeder) {
         return new ResponseEntity<String>(HttpStatus.OK);
     }
 
+    /** Issue a mass truncate request on layers */
+    @RequestMapping(value = "/masstruncate/all", method = RequestMethod.GET)
+    public ResponseEntity<?> doMassTruncateAllGet(HttpServletRequest req) throws IOException {
+        //        String contentType = req.getContentType();
```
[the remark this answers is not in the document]
imranrajjad: will remove

geowebcache/core/src/main/java/org/geowebcache/seed/TruncateAllRequest.java
imranrajjad: will remove and add warning

geowebcache/core/src/main/java/org/geowebcache/seed/TruncateAllRequest.java
imranrajjad: @aaime 
I truncated layers and again seeded them back and forth from UI and REST. After clearing all layers and then asking GWC to seed the layer again went fine both with standalone and integrated GWC. 

So to test your concerns I will follow the below steps:

- Seed some layers
- Nuke the cache
- Seed the layers again and see if there are any issues

Would that do?

aaime: Do the same with disk quota enabled.

geowebcache/rest/src/main/java/org/geowebcache/rest/controller/MassTruncateController.java
```diff
@@ -146,6 +148,45 @@ public void setTileBreeder(TileBreeder breeder) {
         return new ResponseEntity<String>(HttpStatus.OK);
     }
 
+    /** Issue a mass truncate request on layers */
+    @RequestMapping(value = "/masstruncate/all", method = RequestMethod.GET)
```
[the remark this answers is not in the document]
imranrajjad: this method is used by GWC GUI, because I need to response with HTML content. The previous method does not response with anything.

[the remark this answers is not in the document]
aaime: You can fix the old method rather than going off-pattern though?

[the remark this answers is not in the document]
imranrajjad: understood, will fix the old one and remove this one

geowebcache/core/src/main/java/org/geowebcache/seed/TruncateAllRequest.java
imranrajjad: @aaime 

testing with diskquota shows these logs

` (GWC DiskQuota Store Writer-1) INFO [bdb.BDBQuotaStore] - Can't add usage stats. TileSet does not exist. Was it deleted? FX_LP_VT#EPSG:4326#image/png8`

is this related to your concerns?

aaime: Yes, that is a serious problem. Please rewrite the delete code mimicking the geoserver one, deleting by gridset. If you can, please fix TruncateLayer as well.
See: https://github.com/geoserver/geoserver/blob/ac0ed118b8a316de7cd0d8bee6d4d4919ac1e537/src/gwc/src/main/java/org/geoserver/gwc/GWC.java#L356
Then check disk quota works fine, reports zero usage on all layers, and starts collectin stats again as new tiles are getting created.

imranrajjad: will do

imranrajjad:
@aaime implemented requested changes

aaime:
Looks good, merged, looking forward to get the GeoServer one going too

Issue #812, opened 2020-01-05
GWC preview shows blurry tiles


aaime:
The actual tile is not as blurry as it is in the map preview, see here, check the "Azerbaijan" label:

![image](https://user-images.githubusercontent.com/325433/71778410-c473ea00-2fad-11ea-9f8a-46b9c9d9e11c.png)

Seems there is some rescaling involved. Tiles are PNG.

aaime:
Fixed on master, 1.16.x and 1.15.x

PR #813, opened 2020-01-07
GWC preview shows blurry tiles, #812

aaime:
Actually simple fix, see:
https://github.com/openlayers/openlayers/issues/6430
Thanks Stefano Bovio for the hint :-D


aaime:
Will merge during the WE, when I have time to do the backports too.

Issue #817, opened 2020-01-24
RexgexParameterFilter.clone returns a non working clone


aaime:
The cause is that the pattern is not reinitialized, and left to the default "empty string" one.

PR #818, opened 2020-01-24
RexgexParameterFilter.clone returns a non working clone #817

aaime:
@groldan you seem to be the last one to have touched this area of the code, hopefully you remember what was happening. See also https://github.com/GeoWebCache/geowebcache/issues/817

### DIFF
--- a/geowebcache/core/src/main/java/org/geowebcache/filter/parameters/FloatParameterFilter.java
+++ b/geowebcache/core/src/main/java/org/geowebcache/filter/parameters/FloatParameterFilter.java
@@ -147,13 +147,10 @@ public class FloatParameterFilter extends ParameterFilter {
 
     @Override
     public FloatParameterFilter clone() {
-        FloatParameterFilter clone = new FloatParameterFilter();
-        clone.setDefaultValue(getDefaultValue());
-        clone.setKey(getKey());
+        FloatParameterFilter clone = (FloatParameterFilter) super.clone();
         if (values != null) {
-            clone.values = new ArrayList<Float>(values);
+            clone.values = new ArrayList<>(values);
         }
-        clone.setThreshold(this.threshold);
         return clone;
     }
 
--- a/geowebcache/core/src/main/java/org/geowebcache/filter/parameters/IntegerParameterFilter.java
+++ b/geowebcache/core/src/main/java/org/geowebcache/filter/parameters/IntegerParameterFilter.java
@@ -147,13 +147,10 @@ public class IntegerParameterFilter extends ParameterFilter {
 
     @Override
     public IntegerParameterFilter clone() {
-        IntegerParameterFilter clone = new IntegerParameterFilter();
-        clone.setDefaultValue(getDefaultValue());
-        clone.setKey(getKey());
+        IntegerParameterFilter clone = (IntegerParameterFilter) super.clone();
         if (values != null) {
-            clone.values = new ArrayList<Integer>(values);
+            clone.values = new ArrayList<>(values);
         }
-        clone.setThreshold(this.threshold);
         return clone;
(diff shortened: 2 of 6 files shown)


groldan:
Looks like a simple oversight on 989de950bbf50adcdf9db36cbb5be0668c6d62ee, and that the most appropriate patch would be:
```
diff --git a/geowebcache/core/src/main/java/org/geowebcache/filter/parameters/ParameterFilter.java b/geowebcache/core/src/main/java/org/geowebcache/filter/parameters/ParameterFilter.java
index 4d852d9d..20a30e02 100644
--- a/geowebcache/core/src/main/java/org/geowebcache/filter/parameters/ParameterFilter.java
+++ b/geowebcache/core/src/main/java/org/geowebcache/filter/parameters/ParameterFilter.java
@@ -73,7 +73,13 @@ public abstract class ParameterFilter implements Serializable, Cloneable {
     }
 
     @Override
-    public abstract ParameterFilter clone();
+    public ParameterFilter clone() {
+        try {
+            return (ParameterFilter) super.clone();
+        } catch (CloneNotSupportedException e) {
+        }
+    }
 
     /**
      * Checks whether a given parameter value applies to this filter.
diff --git a/geowebcache/core/src/main/java/org/geowebcache/filter/parameters/RegexParameterFilter.java b/geowebcache/core/src/main/java/org/geowebcache/filter/parameters/RegexParameterFilter.java
index 802c59b2..9dff12a2 100644
--- a/geowebcache/core/src/main/java/org/geowebcache/filter/parameters/RegexParameterFilter.java
+++ b/geowebcache/core/src/main/java/org/geowebcache/filter/parameters/RegexParameterFilter.java
@@ -116,10 +116,7 @@ public class RegexParameterFilter extends CaseNormalizingParameterFilter {
 
     @Override
     public RegexParameterFilter clone() {
-        RegexParameterFilter clone = new RegexParameterFilter();
-        clone.setDefaultValue(getDefaultValue());
-        clone.setKey(getKey());
-        clone.regex = regex;
+        RegexParameterFilter clone = (RegexParameterFilter) super.clone();
         if (super.normalize != null) {
             clone.setNormalize(super.normalize.clone());
         }
diff --git a/geowebcache/core/src/test/java/org/geowebcache/filter/parameters/RegexParameterFilterTest.java b/geowebcache/core/src/test/java/org/geowebcache/filter/parameters/RegexParameterFilterTest.java
index d97ee68f..207c4f17 100644
--- a/geowebcache/core/src/test/java/org/geowebcache/filter/parameters/RegexParameterFilterTest.java
+++ b/geowebcache/core/src/test/java/org/geowebcache/filter/parameters/RegexParameterFilterTest.java
@@ -324,5 +324,8 @@ public class RegexParameterFilterTest {
                 clone.normalize.getConfiguredLocale(),
                 equalTo(filter.normalize.getConfiguredLocale()));
         assertThat(clone.normalize.getCase(), equalTo(filter.normalize.getCase()));
+        // actual regex application test
+        assertTrue(filter.applies("Bar"));
+        assertTrue(clone.applies("Bar"));
     }
 }
```

aaime:
Bizzarre... so there was no reason to implement a custom clone back on 2012?

groldan:
it seems to me there was a reason, to avoid shallow clones, yet the implementation could have depended more in the default clone() functionality and only take care of the special cases like this one

aaime:
Ok, reverted the original change, applied the one you suggested, and extended it to the other ParameterFilter subclasses, does not makes sense to have different cloning behaviors for them, same hierarchy, same module and package...

What do you think? (if you decide to merge, let's use a "squash and merge")

groldan:
yes, looks good to me.

PR #824, opened 2020-02-16
Checkstyle maven 3.1.1

aaime:
Updates to latest checkstyle plugin. Since last update, it deprecated and removed some checks, and made javadoc tests stricter (the settings removed were used to make it more lax). Basically had to remove all empty javadoc param/return/throws tags.


smithkm:
It would be nice if it didn't collapse the comments to a single line.

```
/**
  * Blah
  * @return
  */
```
to 
```
/**
  * Blah
  */
```
instead of 
```
/** Blah */
```
but I don't really see any way around that given the formatter

aaime:
Correct, that is done by the formatter, we have no control over it.

PR #827, opened 2020-03-04
Updated HTTPComponents and added lines to specify the version

willettej-leidos:
I updated HTTPComponents and added the necessary lines to specify a particular version to bring GeoWebCache in line with my PRs for both GeoServer and GeoTools.


aaime:
Here we don't have a Jira, release managers here look at the commit history, so no need for a ticket. Merging.

@willettej-leidos can you do a backport on the 1.16.x branch too?

PR #823, opened 2020-02-12
Adding blobstore package which supports Swift

dhlambert:
Extension to GWC which adds native support for Swift. This allows OpenStack Cloud deployments to easily use GeoWebCache without the support from external libraries such as S3Proxy. This has been implemented using the open source multi-cloud toolkit jclouds (https://jclouds.apache.org/).

**Additions:**
- Swift Blobstore package containing blobstore class and supporting config provider classes
- Additions to configuration and pom files for swift
- Unit tests covering blobstore class

PR #852, opened 2020-06-04
swiftblob pom.xml consistency fix

mtnorthcott:
PR to make swiftblob's pom.xml file more consistent with others as described in #823. Moved dependency versions to `geowebcache/pom.xml` and removed redundant sections in `geowebcache/swiftblob/pom.xml`

### DIFF
--- a/geowebcache/pom.xml
+++ b/geowebcache/pom.xml
@@ -50,7 +50,6 @@
     <lint>deprecation</lint>
     <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
     <jclouds.version>2.2.0</jclouds.version>
-    <mockito.version>3.2.4</mockito.version>
   </properties>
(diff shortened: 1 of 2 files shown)


geowebcache/pom.xml
```diff
@@ -49,6 +49,8 @@
     <qa>false</qa>
     <lint>deprecation</lint>
     <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
+    <jclouds.version>2.2.0</jclouds.version>
+    <mockito.version>3.2.4</mockito.version>
```
aaime, line 53: Hmm... probably best not to do this. Mockito version is already used in other modules and managed in this pom, search for mockito-all in this and other poms. However it's an older version of mockito.
If the swift module is doing its own thing best if it's kept private.

PR #857, opened 2020-06-16
Update spring-security to 5.1.11 and spring-framework to 5.1.16

mprins:
See also:
- https://osgeo-org.atlassian.net/browse/GEOS-9650
- https://github.com/geoserver/geoserver/pull/4313


aaime:
Merged thanks

PR #858, opened 2020-06-18
Update the PostgreSQL jdbc driver from 9.4.1208 to 42.2.14

mprins:
Update the PostgreSQL jdbc driver from 9.4.1208 to 42.2.14, resolves CVE-2020-13692 (PostgreSQL JDBC Driver (aka PgJDBC) before 42.2.13 allows XXE) .

see also https://github.com/geoserver/geoserver/pull/4314


aaime:
The downstream build failure seems unrelated, it should not be running against a PostgreSQL database, but let's keep an eye on it. Merging.

PR #866, opened 2020-06-24
Add async functionality to SwiftBlobStore

dhlambert:
Added in async functionality to the `SwiftBlobStore` and updated the tests accordingly. Tiles will now be queued for upload and processed by a dedicated worker pool to drastically reduce upload IO wait.

Some refactoring has also occurred in the put method. In this process, files have been added to the `swiftblob` package including `SwiftTile` and `SwiftUploadTask`

### DIFF
--- a/geowebcache/swiftblob/src/main/java/org/geowebcache/swift/SwiftUploadTask.java
+++ b/geowebcache/swiftblob/src/main/java/org/geowebcache/swift/SwiftUploadTask.java
@@ -49,12 +49,15 @@ public class SwiftUploadTask implements Runnable {
         LocalDateTime time = LocalDateTime.now();
         long getWithoutBody = System.nanoTime();
         SwiftObject object = objectApi.getWithoutBody(key);
-        log.info(
-                String.format(
-                        logStr,
-                        time.format(DateTimeFormatter.ISO_DATE_TIME),
-                        "HEAD",
-                        (System.nanoTime() - getWithoutBody) / 1000000));
+
+        if (log.isDebugEnabled()) {
+            log.debug(
+                    String.format(
+                            logStr,
+                            time.format(DateTimeFormatter.ISO_DATE_TIME),
+                            "HEAD",
+                            (System.nanoTime() - getWithoutBody) / 1000000));
+        }
         if (object == null) {
             return;
         }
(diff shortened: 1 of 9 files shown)


geowebcache/swiftblob/src/main/java/org/geowebcache/swift/SwiftUploadTask.java
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    private void checkExisted() {
+        if (listeners.isEmpty()) {
+            return;
+        }
+        LocalDateTime time = LocalDateTime.now();
+        long getWithoutBody = System.nanoTime();
+        SwiftObject object = objectApi.getWithoutBody(key);
+        log.info(
```
GeoTob, line 52: This should really be a debug or trace level rather than info; same goes for the one in the `run` method below; the point is just to get a rough idea of request/response times during development.

The formatting here is doing __some__ work, so it's also one of those occasions where wrapping it in an `if` statement wouldn't hurt.

geowebcache/swiftblob/src/main/java/org/geowebcache/swift/SwiftUploadTask.java
```diff
@@ -0,0 +1,95 @@
...
+        log.debug("Processing " + key);
+
+        checkExisted();
+
+        try (Payload payload = tile.getPayload()) {
+            String localLogStr = "%s, %s, %dms, %dkB";
+            LocalDateTime time = LocalDateTime.now();
+            long upload = System.nanoTime();
+            objectApi.put(key, payload);
+            log.debug(
```
GeoTob, line 80: See my comment above.

dhlambert:
@jodygarnett @aaime Would it be possible to get a review from you or another maintainer for this pull request?

The PR https://github.com/GeoWebCache/geowebcache/pull/852 has also been updated and needs a re-review also.

geowebcache/swiftblob/src/main/java/org/geowebcache/swift/TMSKeyBuilder.java
```diff
@@ -116,18 +116,16 @@ public String forTile(TileObject obj) {
...
-                        prefix,
-                        layer,
-                        gridset,
-                        shortFormat,
-                        parametersId,
-                        z,
-                        x,
-                        y + "." + extension);
-        return key;
+        return join(
```
aaime, line 119: Please avoid these changes in the future, they are a simple matter of personal preference, we have no fixed rules to choose one way or the other.

aaime:
I can't review the PR, know nothing about Swift, have no use for it, but the changes are all inside the module, and they would not kick in unless the module is used. So there is no reason not to merge!

PR #869, opened 2020-07-09
Add a call to dispose of the ImageReader in MetaTile

dhlambert:
Adds a call to dispose of the `ImageReader` property in `disposePlanarImageChain` method in the class `MetaTile`. 

References: https://osgeo-org.atlassian.net/browse/GEOS-9684


aaime:
Looks good, merging, thanks!
You can backport to 1.17.x and 1.16.x as you see fit.

PR #871, opened 2020-07-22
Updating to jai-ext 1.1.17

aaime:


aaime:
Downstream failure is unrelated, caused by the importer depending on arcsde, which depends on icu, which cannot deal with Java 8 minor versions greater than 255.

Issue #876, opened 2020-08-13
GetCapabilites response for WMTS doesn't contain Service Provider info from geowebcache.xml


ianturton:
When I make a request for GetCapabilties the serviceprovider section comes back as:

```
<ows:ServiceProvider>
<ows:ProviderName>http://localhost:9090/geowebcache/service/wmts</ows:ProviderName>
<ows:ProviderSite xlink:href="http://localhost:9090/geowebcache/service/wmts"/>
<ows:ServiceContact>
<ows:IndividualName>GeoWebCache User</ows:IndividualName>
</ows:ServiceContact>
</ows:ServiceProvider>
```

Despite my having filed in the geowebcache.xml file (which is being read as the layers are correct)! 

Looking at the code the issue seems to be `TileLayerDispatcher` never seems to have it's `serviceInformation` set (in fact there is no setter) - so this is returned as null, and since I don't seem to have any extensions installed when I come to `WMTSGetCapabilities` `getServiceInformation()` can't return anything but Null.

```
private ServiceInformation getServiceInformation() {
        ServiceInformation servInfo = tld.getServiceInformation();
        for (WMTSExtension extension : extensions) {
            ServiceInformation serviceInformation = extension.getServiceInformation();
            if (serviceInformation != null) {
                if (servInfo == null) {
                    servInfo = new ServiceInformation();
                }
                mergeServiceInformation(servInfo, serviceInformation);
            }
        }
        return servInfo;
    }
```

So either I should have an extension? or this has never worked?

ianturton:
See PR https://github.com/GeoWebCache/geowebcache/pull/877

PR #877, opened 2020-08-17
Include the default ServiceInfo into the WMS & WMTS GetCapabilities document

ianturton:
Fix for Issue #876

### DIFF
--- a/geowebcache/core/src/main/java/org/geowebcache/GeoWebCacheDispatcher.java
+++ b/geowebcache/core/src/main/java/org/geowebcache/GeoWebCacheDispatcher.java
@@ -709,14 +709,4 @@ public class GeoWebCacheDispatcher extends AbstractController {
     public void setSecurityDispatcher(SecurityDispatcher secDispatcher) {
         this.securityDispatcher = secDispatcher;
     }
-
-    /** @return the mainConfiguration */
-    public BaseConfiguration getMainConfiguration() {
-        return mainConfiguration;
-    }
-
-    /** @param mainConfiguration the mainConfiguration to set */
-    public void setMainConfiguration(BaseConfiguration mainConfiguration) {
-        this.mainConfiguration = mainConfiguration;
-    }
 }
--- a/geowebcache/core/src/main/java/org/geowebcache/GeoWebCacheDispatcher.java
+++ b/geowebcache/core/src/main/java/org/geowebcache/GeoWebCacheDispatcher.java
@@ -709,4 +709,14 @@ public class GeoWebCacheDispatcher extends AbstractController {
     public void setSecurityDispatcher(SecurityDispatcher secDispatcher) {
         this.securityDispatcher = secDispatcher;
     }
+
+    /** @return the mainConfiguration */
+    public BaseConfiguration getMainConfiguration() {
+        return mainConfiguration;
+    }
+
+    /** @param mainConfiguration the mainConfiguration to set */
+    public void setMainConfiguration(BaseConfiguration mainConfiguration) {
+        this.mainConfiguration = mainConfiguration;
+    }
 }
--- a/geowebcache/core/src/main/java/org/geowebcache/layer/TileLayerDispatcher.java
+++ b/geowebcache/core/src/main/java/org/geowebcache/layer/TileLayerDispatcher.java
@@ -311,7 +311,7 @@ public class TileLayerDispatcher
                         TileLayerConfiguration.class, applicationContext);
 
         XMLConfiguration config = applicationContext.getBean(XMLConfiguration.class);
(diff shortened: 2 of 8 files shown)


geowebcache/core/src/main/java/org/geowebcache/GeoWebCacheDispatcher.java
```diff
@@ -709,4 +709,14 @@ protected SecurityDispatcher getSecurityDispatcher() {
...
+
+    /** @return the mainConfiguration */
+    public BaseConfiguration getMainConfiguration() {
+        return mainConfiguration;
+    }
+
+    /** @param mainConfiguration the mainConfiguration to set */
+    public void setMainConfiguration(BaseConfiguration mainConfiguration) {
+        this.mainConfiguration = mainConfiguration;
+    }
```
smithkm, line 721: This seems to be defeating the purpose of having Spring do dependency injection.  if the Service objects need the Configuration, they should get it from Spring rather that from another object.

geowebcache/wmts/src/main/java/org/geowebcache/service/wmts/WMTSService.java
```diff
@@ -181,11 +182,15 @@ public WMTSService(
...
         this.sb = sb;
         this.tld = tld;
+
         this.gsb = gsb;
         this.stats = stats;
         this.urlMangler = urlMangler;
         this.controller = controller;
         extensions.addAll(GeoWebCacheExtensions.extensions(WMTSExtension.class));
+        if (getMainConfiguration() != null) {
+            tld.setServiceInformation(getMainConfiguration().getServiceInformation());
```
smithkm, line 192: Having the WMTSService constructor responsible for setting the ServiceInformation in the Tile layer dispatcher seems like it's needlessly tying objects together.

geowebcache/wms/src/main/java/org/geowebcache/service/wms/WMSService.java
```diff
@@ -302,7 +302,14 @@ public void handleRequest(Conveyor conv) throws GeoWebCacheException {
...
 
         String servletPrefix = null;
-        if (controller != null) servletPrefix = controller.getServletPrefix();
+
+        if (controller != null) {
+            servletPrefix = controller.getServletPrefix();
+            XMLConfiguration mainConfiguration2 =
+                    (XMLConfiguration) controller.getMainConfiguration();
+            if (mainConfiguration2 != null)
+                tld.setServiceInformation(mainConfiguration2.getServiceInformation());
```
smithkm, line 311: Why is the default service information being set in response to a tile request?

ianturton: I'd be happy for Spring to inject the information into the TLD but I couldn't see any way to make that work, is there an example of that in another class I can look at? 

I'm happy to look at adding a test

smithkm: Well, the TLD is already doing that via two different mechanisms It gets the GridSetBroker as a constructor parameter.  geowebcache-core-context.xml defines the TLD bean so you would then have to update that to tell Spring it needs a reference to the XMLConfiguration bean.

The other method is that the TLD is defined as being ApplicationContextAware which means SPring gives it a refernece to the context, which TLD then uses to extract whatever beans it wants, in this case it looks up all the intances of LayerConfiguration.  Look in the afterPropertiesSet method.

The first method is better practice and is nicer for unit testing but does involve an API change which requires a downstream change in GeoServer (updating the context XML file)  The second method avoids the API change but makes unit tests a bit clunkier as you have to mock the application context.  The second method also requires either coding the id of the BaseConfiguration bean into the TLD, or accounting for the possibility of there being more than one (With the constructor and context files you are telling spring explicitly which bean to use so you don't need to worry about that in the code)

ianturton: OK, I think I've cracked it - I went with the 2nd method so no downstream changes (I hope) but lots of fun with mocking :-)

geowebcache/core/src/main/java/org/geowebcache/layer/TileLayerDispatcher.java
```diff
@@ -302,6 +309,12 @@ public void afterPropertiesSet() throws Exception {
         this.configs =
                 GeoWebCacheExtensions.configurations(
                         TileLayerConfiguration.class, applicationContext);
+
+        XMLConfiguration config = applicationContext.getBean(XMLConfiguration.class);
```
smithkm, line 313: I think this should be BaseConfiguration.class so it can work with alternate implementations like Stratus.

PR #879, opened 2020-09-11
Improve handling of GWC filter parameter storage

jemacchi:
This is the list of features/changes applied based on issue description and talk discussions:

- Update parametersmap (metadata.properties) only during tiles cache folder creation
- Avoid writing metadata file if KVP has no changed
- ParametersMap file compression process
- Minor change on code for naming metadata.properties file
- Optimistic lock approach on metadata file writing/reading
- Improvements based on code review. Backport code. Temporal file creation using File API. Changes in lambda&interface.
- Fix for existing unit tests
- FileBlobStore tests with concurrence on metadata file access/writing

### DIFF
--- a/geowebcache/core/src/main/java/org/geowebcache/storage/blobstore/file/FileBlobStore.java
+++ b/geowebcache/core/src/main/java/org/geowebcache/storage/blobstore/file/FileBlobStore.java
@@ -776,33 +776,22 @@ public class FileBlobStore implements BlobStore {
         final String lockObj = metadataFile.getAbsolutePath().intern();
         synchronized (lockObj) {
             if (metadataFile.exists()) {
-                FileInputStream in;
-                try {
-                    in = new FileInputStream(metadataFile);
+                try (FileInputStream in = new FileInputStream(metadataFile); ) {
+                    int attempts = 0;
+                    while (attempts < FileBlobStore.METADATA_MAX_RW_ATTEMPTS) {
+                        properties.load(in);
+                        // Read current metadatafile date modification (currentModTime)
+                        long currentModDate = metadataFile.lastModified();
+                        if (lastModified == currentModDate) {
+                            break;
+                        } else {
+                            // Try again since some other GWC updated the file
+                            attempts++;
+                            log.debug("Reattempting to read metadata file since timestamp changed");
+                        }
+                    }
                 } catch (IOException e) {
                     throw new UncheckedIOException(e);
-                }
-                int attempts = 0;
-                while (attempts < FileBlobStore.METADATA_MAX_RW_ATTEMPTS) {
-                    try {
-                        properties.load(in);
-                    } catch (Exception e) {
-                        throw new RuntimeException(e);
-                    }
-                    // Read current metadatafile date modification (currentModTime)
-                    long currentModDate = metadataFile.lastModified();
-                    if (lastModified == currentModDate) {
-                        break;
-                    } else {
-                        // Try again since some other GWC updated the file
-                        attempts++;
-                        log.debug("Reattempting to read metadata file since timestamp changed");
-                    }
-                }
-                try {
-                    in.close();
-                } catch (IOException e) {
-                    log.warn(e.getMessage(), e);
                 }
             } else {
                 log.debug("Uncompressed metadata file does not exists: " + metadataFile.getPath());
@@ -817,36 +806,27 @@ public class FileBlobStore implements BlobStore {
         final String lockObj = metadataFile.getAbsolutePath().intern();
         synchronized (lockObj) {
             if (metadataFile.exists()) {
-                FileInputStream in;
-                GZIPInputStream gzipIn;
-                try {
-                    in = new FileInputStream(metadataFile);
-                    gzipIn = new GZIPInputStream(in);
+                try (FileInputStream in = new FileInputStream(metadataFile);
+                        GZIPInputStream gzipIn = new GZIPInputStream(in); ) {
+                    int attempts = 0;
+                    while (attempts < FileBlobStore.METADATA_MAX_RW_ATTEMPTS) {
+                        try {
+                            properties.load(gzipIn);
+                        } catch (Exception e) {
+                            throw new RuntimeException(e);
+                        }
+                        // Read current metadatafile date modification (currentModTime)
+                        long currentModDate = metadataFile.lastModified();
+                        if (lastModified == currentModDate) {
+                            break;
+                        } else {
+                            // Try again since some other GWC updated the file
+                            attempts++;
+                            log.debug("Reattempting to read metadata file since timestamp changed");
+                        }
+                    }
                 } catch (IOException e) {
                     throw new UncheckedIOException(e);
-                }
-                int attempts = 0;
-                while (attempts < FileBlobStore.METADATA_MAX_RW_ATTEMPTS) {
-                    try {
-                        properties.load(gzipIn);
-                    } catch (Exception e) {
-                        throw new RuntimeException(e);
-                    }
-                    // Read current metadatafile date modification (currentModTime)
-                    long currentModDate = metadataFile.lastModified();
-                    if (lastModified == currentModDate) {
-                        break;
-                    } else {
-                        // Try again since some other GWC updated the file
-                        attempts++;
-                        log.debug("Reattempting to read metadata file since timestamp changed");
-                    }
-                }
-                try {
-                    gzipIn.close();
-                    in.close();
-                } catch (IOException e) {
-                    log.warn(e.getMessage(), e);
                 }
             } else {
                 log.debug("Metadata file does not exists: " + metadataFile.getPath());
--- a/geowebcache/core/src/main/java/org/geowebcache/storage/blobstore/file/FileBlobStore.java
+++ b/geowebcache/core/src/main/java/org/geowebcache/storage/blobstore/file/FileBlobStore.java
@@ -26,8 +26,10 @@ import java.io.FileNotFoundException;
 import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.OutputStream;
+import java.io.OutputStreamWriter;
 import java.io.UncheckedIOException;
 import java.io.UnsupportedEncodingException;
+import java.io.Writer;
 import java.net.URLDecoder;
 import java.net.URLEncoder;
 import java.nio.channels.FileChannel;
@@ -47,6 +49,8 @@ import java.util.concurrent.atomic.AtomicLong;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import java.util.stream.StreamSupport;
+import java.util.zip.GZIPInputStream;
+import java.util.zip.GZIPOutputStream;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 import org.geowebcache.GeoWebCacheException;
@@ -75,6 +79,10 @@ public class FileBlobStore implements BlobStore {
             LogFactory.getLog(org.geowebcache.storage.blobstore.file.FileBlobStore.class);
 
     static final int DEFAULT_DISK_BLOCK_SIZE = 4096;
+
+    static final String METADATA_GZIP_EXTENSION = ".gz";
+
+    static final int METADATA_MAX_RW_ATTEMPTS = 10;
 
     public static final int BUFFER_SIZE = 32768;
 
@@ -364,7 +372,7 @@ public class FileBlobStore implements BlobStore {
 
     /** Delete a particular tile */
     public boolean delete(TileObject stObj) throws StorageException {
-        File fh = getFileHandleTile(stObj, false);
+        File fh = getFileHandleTile(stObj, null);
         boolean ret = false;
         // we call fh.length() here to check wthether the file exists and its length in a single
         // operation cause lots of calls to exists() may raise the file system cache usage to the
@@ -456,7 +464,7 @@ public class FileBlobStore implements BlobStore {
      * @return true if successful, false otherwise
      */
     public boolean get(TileObject stObj) throws StorageException {
-        File fh = getFileHandleTile(stObj, false);
+        File fh = getFileHandleTile(stObj, null);
         if (!fh.exists()) {
             stObj.setStatus(Status.MISS);
             return false;
@@ -471,9 +479,16 @@ public class FileBlobStore implements BlobStore {
 
     /** Store a tile. */
     public void put(TileObject stObj) throws StorageException {
-        final File fh = getFileHandleTile(stObj, true);
+
+        // An update to ParameterMap file is required !!!
+        Runnable upm =
+                () -> {
+                    this.persistParameterMap(stObj);
+                };
+        final File fh = getFileHandleTile(stObj, upm);
         final long oldSize = fh.length();
         final boolean existed = oldSize > 0;
+
         writeFile(fh, stObj, existed);
 
         // mark the last modification as the tile creation time if set, otherwise
@@ -497,7 +512,7 @@ public class FileBlobStore implements BlobStore {
         }
     }
 
-    private File getFileHandleTile(TileObject stObj, boolean create) throws StorageException {
+    private File getFileHandleTile(TileObject stObj, Runnable upmFunction) throws StorageException {
         final MimeType mimeType;
         try {
             mimeType = MimeType.createFromFormat(stObj.getBlobFormat());
@@ -513,9 +528,12 @@ public class FileBlobStore implements BlobStore {
             throw new StorageException("Failed to compute file path", e);
         }
 
-        if (create) {
+        // If an update to ParameterMap is required, then folder has to be created
+        if (upmFunction != null) {
+            log.debug("Creating parent tile folder and updating ParameterMap");
             File parent = tilePath.getParentFile();
             mkdirs(parent, stObj);
+            upmFunction.run();
         }
 
         return tilePath;
@@ -554,7 +572,6 @@ public class FileBlobStore implements BlobStore {
                 }
             }
 
-            persistParameterMap(stObj);
         } finally {
 
             if (temp != null) {
@@ -644,27 +661,104 @@ public class FileBlobStore implements BlobStore {
      *     java.lang.String)
      */
     public void putLayerMetadata(final String layerName, final String key, final String value) {
-        Properties metadata = getLayerMetadata(layerName);
+        // Get current metadataFile
+        final File metadataFile = getMetadataFile(layerName);
+        // Get last modification time for metadata file
+        long lastModDate = metadataFile.lastModified();
+        // Get metadata content from current metadataFile
+        Properties metadata = getLayerMetadata(metadataFile);
+
+        // Evaluate updating key arg related values (remove/add/update or skip)
+        boolean skipUpdate = false;
+        // Note: since we want to avoid writing metadata file if nothing changed, then we need to
+        // check if it's required
+        // or not (flag skipUpdate). That is reason not to use metadata.merge method instead of
+        // using setProperty
         if (null == value) {
             metadata.remove(key);
         } else {
+            String currValue = metadata.getProperty(key);
+            String encodedValue = null;
             try {
-                metadata.setProperty(key, URLEncoder.encode(value, "UTF-8"));
+                encodedValue = URLEncoder.encode(value, "UTF-8");
             } catch (UnsupportedEncodingException e) {
                 throw new RuntimeException(e);
             }
+            skipUpdate =
+                    currValue != null && encodedValue != null && currValue.equals(encodedValue);
+            // If existing value for key is the same as value arg, then avoid to write it
+            if (!skipUpdate) {
+                metadata.setProperty(key, encodedValue);
+            } else {
+                skipUpdate = true;
+            }
         }
 
-        final File metadataFile = getMetadataFile(layerName);
+        // We call to save metadata only if has changed
+        if (!skipUpdate) {
+            // Write metadata in a temp file
+            File tempFile = null;
+            try {
+                tempFile = writeTempMetadataFile(metadata);
+            } catch (IOException e) {
+                log.error("Cannot create temporary file for " + metadataFile.getPath());
+                throw new UncheckedIOException(e);
+            }
+            // Attempt to rename tmp file to metadata file
+            int tryagain = 0;
+            while (tryagain < FileBlobStore.METADATA_MAX_RW_ATTEMPTS) {
+                // Read current metadatafile date modification (currentModTime)
+                long currentModDate = metadataFile.lastModified();
+                if (lastModDate == 0 || lastModDate == currentModDate) {
+                    if (!metadataFile.getParentFile().exists()) {
+                        metadataFile.getParentFile().mkdirs();
+                    }
+                    // atomic rename of temporary metadata file with getMetadataFilename
+                    // file, in such case we'll just eliminate this one
+                    if (FileUtils.renameFile(tempFile, metadataFile)) {
+                        tempFile = null;
+                    } else {
+                        // Renaming failed !
+                        log.error("Error while renaming metadata file " + metadataFile.getPath());
+                    }
+                    tryagain = FileBlobStore.METADATA_MAX_RW_ATTEMPTS;
+                } else {
+                    // Try again since some other GWC updated the file in middle of writing process
+                    ++tryagain;
+                    log.debug("Reattempting to write metadata file since timestamp changed");
+                }
+            }
+        }
+    }
 
+    private File writeTempMetadataFile(Properties metadata) throws IOException {
+        tmp.mkdirs();
+        final File metadataFile =
+                File.createTempFile("tmp", FileBlobStore.METADATA_GZIP_EXTENSION, tmp);
+        return this.writeMetadataFile(metadata, metadataFile);
+    }
+
+    /**
+     * Writes a Metadatafile with metadata parameter content
+     *
+     * @param metadata
+     * @return temporal file or null if failed
+     * @throws IOException
+     */
+    private File writeMetadataFile(Properties metadata, File metadataFile) throws IOException {
         final String lockObj = metadataFile.getAbsolutePath().intern();
         synchronized (lockObj) {
             if (!metadataFile.getParentFile().exists()) {
                 metadataFile.getParentFile().mkdirs();
             }
-            try (OutputStream out = new FileOutputStream(metadataFile)) {
+            try (OutputStream out = new FileOutputStream(metadataFile);
+                    Writer writer = new OutputStreamWriter(new GZIPOutputStream(out), "UTF-8"); ) {
+
                 String comments = "auto generated file, do not edit by hand";
-                metadata.store(out, comments);
+                // Save as a gzip file
+                metadata.store(writer, comments);
+                return metadataFile;
+
             } catch (FileNotFoundException e) {
                 throw new UncheckedIOException(e);
             } catch (Exception e) {
@@ -673,8 +767,7 @@ public class FileBlobStore implements BlobStore {
         }
     }
 
-    private Properties getLayerMetadata(final String layerName) {
-        final File metadataFile = getMetadataFile(layerName);
+    private Properties getUncompressedLayerMetadata(final File metadataFile) {
         Properties properties = new Properties();
         final String lockObj = metadataFile.getAbsolutePath().intern();
         synchronized (lockObj) {
@@ -682,7 +775,7 @@ public class FileBlobStore implements BlobStore {
                 FileInputStream in;
                 try {
                     in = new FileInputStream(metadataFile);
-                } catch (FileNotFoundException e) {
+                } catch (IOException e) {
                     throw new UncheckedIOException(e);
                 }
                 try {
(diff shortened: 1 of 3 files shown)


aaime:
Legit failure in QA checks:

```
[ERROR] Failed to execute goal com.coveo:fmt-maven-plugin:2.4.0:check (default) on project gwc-core: Found 2 non-complying files, failing build -> [Help 1]
```

All you need to fix formatting, is to build using maven on the command line, it will reformat all files according to specification. You can read more about it here:
http://docs.geotools.org/latest/developer/conventions/code/style.html
Formatting is just the first of the QA checks, you can learn about the others (and how to run them on your computer) here:
http://docs.geotools.org/latest/developer/conventions/code/qa.html

Links are for GeoTools, but all 3 projects, GeoTools, GeoWebCache and GeoServer, use the same checks.

geowebcache/core/src/main/java/org/geowebcache/storage/blobstore/file/FileBlobStore.java
```diff
@@ -497,7 +512,7 @@ public void put(TileObject stObj) throws StorageException {
         }
     }
 
-    private File getFileHandleTile(TileObject stObj, boolean create) throws StorageException {
+    private File getFileHandleTile(TileObject stObj, Runnable upmFunction) throws StorageException {
```
aaime, line 515: You are using it to update the parameter map, but in the context of this method, it's just something happening as the parent tile folder is created (could be anything depending on the needs of the caller). I'd call it "onTileFolderCreation" instead.

geowebcache/core/src/main/java/org/geowebcache/storage/blobstore/file/FileBlobStore.java
```diff
@@ -513,9 +528,12 @@ private File getFileHandleTile(TileObject stObj, boolean create) throws StorageE
             throw new StorageException("Failed to compute file path", e);
         }
 
-        if (create) {
+        // If an update to ParameterMap is required, then folder has to be created
```
aaime, line 531: Update the comment to be more generic.

geowebcache/core/src/main/java/org/geowebcache/storage/blobstore/file/FileBlobStore.java
```diff
@@ -644,27 +661,104 @@ private static String urlDecUtf8(String value) {
...
+        final File metadataFile = getMetadataFile(layerName);
+        // Get last modification time for metadata file
+        long lastModDate = metadataFile.lastModified();
+        // Get metadata content from current metadataFile
+        Properties metadata = getLayerMetadata(metadataFile);
+
+        // Evaluate updating key arg related values (remove/add/update or skip)
+        boolean skipUpdate = false;
+        // Note: since we want to avoid writing metadata file if nothing changed, then we need to
+        // check if it's required
```
aaime, line 674: Adjust comments not to be broken on so many lines.

geowebcache/core/src/main/java/org/geowebcache/storage/blobstore/file/FileBlobStore.java
```diff
@@ -644,27 +661,104 @@ private static String urlDecUtf8(String value) {
...
             } catch (UnsupportedEncodingException e) {
                 throw new RuntimeException(e);
             }
+            skipUpdate =
+                    currValue != null && encodedValue != null && currValue.equals(encodedValue);
+            // If existing value for key is the same as value arg, then avoid to write it
+            if (!skipUpdate) {
+                metadata.setProperty(key, encodedValue);
+            } else {
+                skipUpdate = true;
```
aaime, line 693: Redundant, can get here only if skipUpdate is already true.

geowebcache/core/src/main/java/org/geowebcache/storage/blobstore/file/FileBlobStore.java
```diff
@@ -644,27 +661,104 @@ private static String urlDecUtf8(String value) {
...
+                throw new UncheckedIOException(e);
+            }
+            // Attempt to rename tmp file to metadata file
+            int tryagain = 0;
+            while (tryagain < FileBlobStore.METADATA_MAX_RW_ATTEMPTS) {
+                // Read current metadatafile date modification (currentModTime)
+                long currentModDate = metadataFile.lastModified();
+                if (lastModDate == 0 || lastModDate == currentModDate) {
+                    if (!metadataFile.getParentFile().exists()) {
+                        metadataFile.getParentFile().mkdirs();
```
aaime, line 714: mkdirs does not need to be guarded, it already creates directories only if needed.

geowebcache/core/src/main/java/org/geowebcache/storage/blobstore/file/FileBlobStore.java
```diff
@@ -644,27 +661,104 @@ private static String urlDecUtf8(String value) {
...
+                    }
+                    // atomic rename of temporary metadata file with getMetadataFilename
+                    // file, in such case we'll just eliminate this one
+                    if (FileUtils.renameFile(tempFile, metadataFile)) {
+                        tempFile = null;
+                    } else {
+                        // Renaming failed !
+                        log.error("Error while renaming metadata file " + metadataFile.getPath());
+                    }
+                    tryagain = FileBlobStore.METADATA_MAX_RW_ATTEMPTS;
```
aaime, line 724: If you want to immediately exit, using a "break" or a "return" is clearer than forcing the exit condition of the loop.

geowebcache/core/src/main/java/org/geowebcache/storage/blobstore/file/FileBlobStore.java
```diff
@@ -644,27 +661,104 @@ private static String urlDecUtf8(String value) {
...
+            } else {
+                skipUpdate = true;
+            }
         }
 
-        final File metadataFile = getMetadataFile(layerName);
+        // We call to save metadata only if has changed
+        if (!skipUpdate) {
+            // Write metadata in a temp file
+            File tempFile = null;
```
aaime, line 700: This entire block could be in its own method, "writeOptimisticLock", the name would also give away what it is intending to do.

geowebcache/core/src/main/java/org/geowebcache/storage/blobstore/file/FileBlobStore.java
```diff
@@ -644,27 +661,104 @@ private static String urlDecUtf8(String value) {
...
+            } catch (IOException e) {
+                log.error("Cannot create temporary file for " + metadataFile.getPath());
+                throw new UncheckedIOException(e);
+            }
+            // Attempt to rename tmp file to metadata file
+            int tryagain = 0;
+            while (tryagain < FileBlobStore.METADATA_MAX_RW_ATTEMPTS) {
+                // Read current metadatafile date modification (currentModTime)
+                long currentModDate = metadataFile.lastModified();
+                if (lastModDate == 0 || lastModDate == currentModDate) {
```
aaime, line 712: How do we get a lastModeDate == 0? (besides someone having the system clock set to the seventies?)

geowebcache/core/src/main/java/org/geowebcache/storage/blobstore/file/FileBlobStore.java
```diff
@@ -644,27 +661,104 @@ private static String urlDecUtf8(String value) {
...
+                    if (!metadataFile.getParentFile().exists()) {
+                        metadataFile.getParentFile().mkdirs();
+                    }
+                    // atomic rename of temporary metadata file with getMetadataFilename
+                    // file, in such case we'll just eliminate this one
+                    if (FileUtils.renameFile(tempFile, metadataFile)) {
+                        tempFile = null;
```
aaime, line 719: No need to set local variables to null, could be a good place to just exit.

geowebcache/core/src/main/java/org/geowebcache/storage/blobstore/file/FileBlobStore.java
```diff
@@ -644,27 +661,104 @@ private static String urlDecUtf8(String value) {
...
+                    } else {
+                        // Renaming failed !
+                        log.error("Error while renaming metadata file " + metadataFile.getPath());
```
aaime, line 722: With Windows as a possible target, the rename could be failing due to a conflict... I would make it retry in this conditionnn.

geowebcache/core/src/main/java/org/geowebcache/storage/blobstore/file/FileBlobStore.java
```diff
@@ -644,27 +661,104 @@ private static String urlDecUtf8(String value) {
...
+                    }
+                    tryagain = FileBlobStore.METADATA_MAX_RW_ATTEMPTS;
+                } else {
+                    // Try again since some other GWC updated the file in middle of writing process
+                    ++tryagain;
```
aaime, line 727: Not wrong, but a bit out of place, the idiomatic way to perform a stand alone counter increment is "tryagain++"

geowebcache/core/src/main/java/org/geowebcache/storage/blobstore/file/FileBlobStore.java
```diff
@@ -644,27 +661,104 @@ private static String urlDecUtf8(String value) {
...
+            // Write metadata in a temp file
+            File tempFile = null;
+            try {
+                tempFile = writeTempMetadataFile(metadata);
+            } catch (IOException e) {
+                log.error("Cannot create temporary file for " + metadataFile.getPath());
+                throw new UncheckedIOException(e);
+            }
+            // Attempt to rename tmp file to metadata file
+            int tryagain = 0;
```
aaime, line 708: Variable name suggest a boolean "should I try again", but the usage is the one of a counter. Suggest calling it "tries" or "attempts" instead.

geowebcache/core/src/main/java/org/geowebcache/storage/blobstore/file/FileBlobStore.java
```diff
@@ -696,15 +789,108 @@ private Properties getLayerMetadata(final String layerName) {
                         log.warn(e.getMessage(), e);
                     }
                 }
+            } else {
+                log.error(
```
aaime, line 793: Not finding a metadata file is a normal condition, maybe filter parameters were never used before. Why log it at error level?

geowebcache/core/src/main/java/org/geowebcache/storage/blobstore/file/FileBlobStore.java
```diff
@@ -696,15 +789,108 @@ private Properties getLayerMetadata(final String layerName) {
...
+                FileInputStream in;
+                GZIPInputStream gzipIn;
+                try {
+                    in = new FileInputStream(metadataFile);
+                    gzipIn = new GZIPInputStream(in);
+                } catch (IOException e) {
+                    throw new UncheckedIOException(e);
+                }
+                try {
+                	int tryagain = 0;
```
aaime, line 815: Same observation as above.

geowebcache/core/src/main/java/org/geowebcache/storage/blobstore/file/FileBlobStore.java
```diff
@@ -696,15 +789,108 @@ private Properties getLayerMetadata(final String layerName) {
...
+                    while (tryagain < FileBlobStore.METADATA_MAX_RW_ATTEMPTS) {
+                        // Read current metadatafile date modification (currentModTime)
+                        long currentModDate = metadataFile.lastModified();
+	                	if (lastModified == currentModDate) {
+	                		properties.load(gzipIn);
+	                		tryagain = FileBlobStore.METADATA_MAX_RW_ATTEMPTS;
```
aaime, line 821: Same observation as above.

geowebcache/core/src/main/java/org/geowebcache/storage/blobstore/file/FileBlobStore.java
```diff
@@ -696,15 +789,108 @@ private Properties getLayerMetadata(final String layerName) {
...
+                    gzipIn = new GZIPInputStream(in);
+                } catch (IOException e) {
+                    throw new UncheckedIOException(e);
+                }
+                try {
+                	int tryagain = 0;
+                    while (tryagain < FileBlobStore.METADATA_MAX_RW_ATTEMPTS) {
+                        // Read current metadatafile date modification (currentModTime)
+                        long currentModDate = metadataFile.lastModified();
+	                	if (lastModified == currentModDate) {
```
aaime, line 819: It's safer to first read and then check if the last modified date changed, ensures that nobody modified the file during the read.

geowebcache/core/src/main/java/org/geowebcache/storage/blobstore/file/FileBlobStore.java
```diff
@@ -696,15 +789,108 @@ private Properties getLayerMetadata(final String layerName) {
...
+	                        // Try again since some other GWC updated the file in middle of writing process
+	                        ++tryagain;
+	                        log.debug("Reattempting to read metadata file since timestamp changed");
+	                    }
+                    }
+                    
+                } catch (Exception e) {
+                    throw new RuntimeException(e);
+                } finally {
+                    try {
```
aaime, line 832: Better to use try-with-resources, just use the construct inside the retry loop.

geowebcache/core/src/main/java/org/geowebcache/storage/blobstore/file/FileBlobStore.java
```diff
@@ -696,15 +789,108 @@ private Properties getLayerMetadata(final String layerName) {
...
+                        gzipIn.close();
+                        in.close();
+                    } catch (IOException e) {
+                        log.warn(e.getMessage(), e);
+                    }
+                }
+            } else {
+                log.error("Metadata file does not exists: " + metadataFile.getPath());
```
aaime, line 840: See above, not an error, should be logged at most at debug level.

geowebcache/core/src/main/java/org/geowebcache/storage/blobstore/file/FileBlobStore.java
```diff
@@ -696,15 +789,108 @@ private Properties getLayerMetadata(final String layerName) {
...
+     * @return metadata file (compressed or not, depending if it's present uncompressed)
+     */
     private File getMetadataFile(final String layerName) {
         File layerPath = getLayerPath(layerName);
-        File metadataFile = new File(layerPath, "metadata.properties");
-        return metadataFile;
+        String metadataFilename = this.getMetadataFilename(true);
+        File metadataFile = new File(layerPath, metadataFilename);
+
+        // At this point we backport uncompressed metadata.properties files to compressed version
```
aaime, line 870: This is not a backport, it's for "backwards compatibility".

aaime: Also, I'd first check if the (compressed) metadata file exists, which would be the new long term "normality", and only when it's missing, see if there is an older metadata file to read and convert.

geowebcache/core/src/main/java/org/geowebcache/storage/blobstore/file/FileBlobStore.java
```diff
@@ -673,16 +767,15 @@ public void putLayerMetadata(final String layerName, final String key, final Str
         }
     }
 
-    private Properties getLayerMetadata(final String layerName) {
-        final File metadataFile = getMetadataFile(layerName);
+    private Properties getUncompressedLayerMetadata(final File metadataFile) {
```
aaime, line 770: Multiple thread/processes could be trying to migrate the old property file too. Suggest using optimistic checks here too, also consider that during migration the file can disappear, if the other "actor" in place manages to do the migration before we do.

geowebcache/core/src/test/java/org/geowebcache/blobstore/file/FileBlobStoreComformanceTest.java
```diff
@@ -27,4 +35,69 @@
...
+                        latch.countDown();
+                    });
+        }
+        latch.await();
+        assertThat(store.getLayerMetadata("testLayer", "test.Key.1"), equalTo("testValue"));
+        assertThat(store.getLayerMetadata("testLayer", "test.Key.0"), equalTo("testValue"));
+    }
+
+    @Test
+    public void testConcurrentMassiveMetadataKeys() throws InterruptedException {
```
aaime, line 73: This could take a long time, useful to have for development and eventual double checks, but I'd mark it as "@Ignore" long term, or at least make sure it does not take too much time (seconds is fine, minutes is not) and cannot slam down the machine (50 threads on a single core computer is too much, on a 128 cores one is not enough, what about using Runtime to get the number of cores, and make the test use 2-4 times more than available?).

aaime (changes requested):

jemacchi:
@aaime  all suggestions were included.

geowebcache/core/src/main/java/org/geowebcache/storage/blobstore/file/FileBlobStore.java
aaime: Still seems not to be using try-with-resources.

geowebcache/core/src/main/java/org/geowebcache/storage/blobstore/file/FileBlobStore.java
aaime: The position of the comment was probably a bit confusing. If someone else manages to do the migration before this code does, and the file disappears, it should check the compressed metata file.

aaime (commented):
Almost there, still a couple of things a bit off.
Also, the QA tests are still failing, this time on PMD:

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-pmd-plugin:3.11.0:check (default) on project gwc-core: You have 3 PMD violations. For more details see: /home/travis/build/GeoWebCache/geowebcache/geowebcache/core/target/pmd.xml -> [Help 1]
```

The [link I have you earlier](http://docs.geotools.org/latest/developer/conventions/code/qa.html) shows how to run the full set of QA tools on your own machine. 
Failing that, you can also check the build results, as part of the Travis build:

image

image

image

jemacchi (commented):
> Almost there, still a couple of things a bit off.
> Also, the QA tests are still failing, this time on PMD:
> 
> ```
> [ERROR] Failed to execute goal org.apache.maven.plugins:maven-pmd-plugin:3.11.0:check (default) on project gwc-core: You have 3 PMD violations. For more details see: /home/travis/build/GeoWebCache/geowebcache/geowebcache/core/target/pmd.xml -> [Help 1]
> ```
> 
> The [link I have you earlier](http://docs.geotools.org/latest/developer/conventions/code/qa.html) shows how to run the full set of QA tools on your own machine.
> Failing that, you can also check the build results, as part of the Travis build:
> 
> image
> 
> image
> 
> image
> 
> I like it better to check on my machine, but you can also see the errors in the logs, e.g.:
> 
> ```
> [INFO] --- maven-pmd-plugin:3.11.0:check (default) @ gwc-core ---
> [INFO] PMD Failure: org.geowebcache.storage.blobstore.file.FileBlobStore:779 Rule:CloseResource Priority:3 Ensure that resources like this FileInputStream object are closed after use.
> [INFO] PMD Failure: org.geowebcache.storage.blobstore.file.FileBlobStore:820 Rule:CloseResource Priority:3 Ensure that resources like this FileInputStream object are closed after use.
> [INFO] PMD Failure: org.geowebcache.storage.blobstore.file.FileBlobStore:821 Rule:CloseResource Priority:3 Ensure that resources like this GZIPInputStream object are closed after use.
> [INFO] PMD Failure: org.geowebcache.storage.blobstore.file.FileBlobStore:779 Rule:CloseResource Priority:3 Ensure that resources like this FileInputStream object are closed after use.
> [INFO] PMD Failure: org.geowebcache.storage.blobstore.file.FileBlobStore:820 Rule:CloseResource Priority:3 Ensure that resources like this FileInputStream object are closed after use.
> [INFO] PMD Failure: org.geowebcache.storage.blobstore.file.FileBlobStore:821 Rule:CloseResource Priority:3 Ensure that resources like this GZIPInputStream object are closed after use.
> ```
> 
> Believe it's the try-with-resources issue I've commented about above.

Yes, you are right, my bad about understanding how it's the QA option working.

geowebcache/core/src/test/java/org/geowebcache/blobstore/file/FileBlobStoreComformanceTest.java
```diff
@@ -27,4 +36,103 @@
...
+
+    @Test
+    public void testMetadataWithPointInKey() throws Exception {
+        assertThat(store.getLayerMetadata("testLayer", "test.Key"), nullValue());
+        store.putLayerMetadata("testLayer", "test.Key", "testValue");
+        assertThat(store.getLayerMetadata("testLayer", "test.Key"), equalTo("testValue"));
+    }
+
+    @Test
+    public void testConcurrentMetadataWithPointInKey() throws InterruptedException {
```
nmco, line 48: Looks like these tests share a lot of code, could it be possible to create a support method contains the common code?

jemacchi:
@nmco + @aaime  your comments/suggestions were included in last commit. Please take a look when possible.

aaime:
The changes look good to me... and yet, it seems there is still enough of a race contidion window to have tests fail. This comes from the failed OSX build:

```
2020-09-15T19:36:08.7413160Z [ERROR] testConcurrentMetadataBasedOnCores(org.geowebcache.blobstore.file.FileBlobStoreComformanceTest)  Time elapsed: 1.862 s  <<< FAILURE!
2020-09-15T19:36:08.7416240Z java.lang.AssertionError: 
2020-09-15T19:36:08.7418620Z 
2020-09-15T19:36:08.7419110Z Expected: "testValue4"
2020-09-15T19:36:08.7419750Z      but: was null
2020-09-15T19:36:08.7422170Z 	at org.geowebcache.blobstore.file.FileBlobStoreComformanceTest.testConcurrentMetadataBasedOnCores(FileBlobStoreComformanceTest.java:107)
```

jemacchi:
Yes, it seems that the approach for testing based on multiples threads
competing it's not going to work consistently (also cause number of
attempts to write/read in an optimistic way is a fixed number)
I think that I will need to add a different approach to test concurrency
(ie. forcing a manual change on lastModified data associated to metadata
file in middle of a putLayerMetadata call, or something like that)

aaime:
Having a flacky test is indeed a problem. But this failure shows there are still concurrency issues. Those need to be solved.

aaime:
Nice, all checks are passing:

image

Also good call to isolate the metadata handling in a separate class. Having a look at it.

geowebcache/core/src/main/java/org/geowebcache/storage/blobstore/file/LayerMetadataStore.java
```diff
@@ -0,0 +1,358 @@
...
+
+    private final String path;
+
+    private File tmp;
+
+    /** number of locks, make it configurable maybe */
+    private static final int lockShardSize = 32;
+
+    /** handling of local-process concurrent access to layer metadata files */
+    private ReadWriteLock[] locks =
```
aaime, line 69: This helps in single instance, especially under load, but won't do a thing when using multiple GWC instances, on different nodes, concurrently accessing the same storage.

geowebcache/core/src/main/java/org/geowebcache/storage/blobstore/file/LayerMetadataStore.java
```diff
@@ -0,0 +1,358 @@
...
+    private ReadWriteLock getLock(File file) {
+        return locks[resolveLockBucket(file)];
+    }
+
+    /**
+     * Performs the actual update of the metatada, making sure only the provided key/value pair is
+     * updated in case another process modified the metadata file since it was loaded by the caller
+     * code
+     */
+    @SuppressFBWarnings(value = "DLS_DEAD_LOCAL_STORE")
```
aaime, line 151: Which dead local store was suppressed here?

jemacchi:
Line 183 in LayerMetadataStore triggered when option -Dqa enabled
metadata = loadLayerMetadata(metadataFile);

jemacchi:
@aaime 
It were included new tests that are basically creating new FileBlobStores and executing updates on same layer. This emulates behavior of concurrent GWC instances in a distributed system (those tests are Ignored currently, since in some cases Optimistic lock will fail depending on system conditions). 
Also it was included the same re-attempt strategy based on differences on content saved during metadata file renaming (atomic writing). If differs then it's required to reattempt.

geowebcache/core/src/main/java/org/geowebcache/storage/blobstore/file/LayerMetadataStore.java
```diff
@@ -0,0 +1,401 @@
...
+
+        log.debug("Start attempt to add key (key: " + key + ")");
+
+        rwLock.writeLock().lock();
+        try {
+            createParentIfNeeded(metadataFile);
+            int attempt = 0;
+            for (attempt = 0; attempt < maxAttempts; attempt++) {
+                if (lastModified == metadataFile.lastModified()) {
+                    metadata = loadLayerMetadata(metadataFile);
```
aaime, line 172: Since we rewrite the metadata object reference here, what's the point of passing them as a parameter? I don't see the value being used before.

aaime:
Tried out the test with the multiple blob stores, it fails every single time on my machine (Ryzen 1700x, 16 cores, thus, 16 threads in the test).

Checking runs with some debug, I believe two writes can conflict on the atomic rename + subsequente test (both running without interruption, and then another thread doing the rename and test), and when it happens, well, it's too late to recover.

This fixes is most of the time:

```
diff --git a/geowebcache/core/src/main/java/org/geowebcache/storage/blobstore/file/LayerMetadataStore.java b/geowebcache/core/src/main/java/org/geowebcache/storage/blobstore/file/LayerMetadataStore.java
index 875dc742..a49bbbf1 100644
--- a/geowebcache/core/src/main/java/org/geowebcache/storage/blobstore/file/LayerMetadataStore.java
+++ b/geowebcache/core/src/main/java/org/geowebcache/storage/blobstore/file/LayerMetadataStore.java
@@ -172,27 +172,32 @@ public class LayerMetadataStore {
                     metadata = loadLayerMetadata(metadataFile);
                     metadata.compute(key, (k, oldValue) -> value); // removes mapping if value==null
                     File tempFile = writeTempMetadataFile(metadata);
-                    if (FileUtils.renameFile(tempFile, metadataFile)) {
-                        // compare content between renamed file and memory content
-                        Properties metadataAfterRename = loadLayerMetadata(metadataFile);
-                        if (!metadata.equals(metadataAfterRename)) {
-                            log.debug(
-                                    "Renamed file content differs from expected saved content.\nCurrent:"
-                                            + metadataAfterRename.toString()
-                                            + "\nExpected: "
-                                            + metadata.toString());
-                            attempt++;
+                    if (lastModified == metadataFile.lastModified()) {
+                        if (FileUtils.renameFile(tempFile, metadataFile)) {
+                            Thread.sleep(50);
+                            // compare content between renamed file and memory content
+                            Properties metadataAfterRename = loadLayerMetadata(metadataFile);
+                            if (!metadata.equals(metadataAfterRename)) {
+                                log.debug(
+                                        "Renamed file content differs from expected saved content.\nCurrent:"
+                                                + metadataAfterRename.toString()
+                                                + "\nExpected: "
+                                                + metadata.toString());
+                                attempt++;
+                            } else {
+                                log.debug(
+                                        "Temporary file renamed successfully (metadata: "
+                                                + metadata.toString()
+                                                + ")");
+                                return;
+                            }
                         } else {
-                            log.debug(
-                                    "Temporary file renamed successfully (metadata: "
-                                            + metadata.toString()
-                                            + ")");
-                            return;
+                            log.info(
+                                    "Reattempting to write metadata file, because an error while renaming metadata file "
+                                            + metadataFile.getPath());
                         }
                     } else {
-                        log.info(
-                                "Reattempting to write metadata file, because an error while renaming metadata file "
-                                        + metadataFile.getPath());
+                        attempt++;
                     }
                     tempFile.delete();
                 } else {
@@ -216,6 +221,8 @@ public class LayerMetadataStore {
             if (maxAttempts == attempt) {
                 log.debug("Optimistic write reaches max number of attempts (" + maxAttempts + ")");
             }
+        } catch (InterruptedException e) {
         } finally {
             rwLock.writeLock().unlock();
         }
@@ -258,6 +265,7 @@ public class LayerMetadataStore {
     private void createParentIfNeeded(File metadataFile) {
         File parentDir = metadataFile.getParentFile();
         if (!parentDir.exists() && !parentDir.mkdirs()) {
+            if (!parentDir.exists())
             throw new IllegalStateException(
                     "Unable to create parent directory " + parentDir.getAbsolutePath());
         }
```

The if per se is more safety. The really sore eye there is the "Thread.sleep". It allows to stop the current thread, allow other threads to complete their rename, and check that a conflict actually happened. I've tried with something a bit nicer, 
 ``Thread.yield()``, but it does nothing. Maybe the wait time should be made configurable too (sysvar with default value).

jemacchi:
@aaime  
Based on test I ran, while debugging, specifically on those cases when more than one FileBlobStore is trying to write what is causing the failure of those concurrent test cases is reaching the PROPERTY_METADATA_MAX_RW_ATTEMPTS  on cases of reading (then operation is discarded)
So the approach you are considering is right (giving some time to let threads knows that something else is happening). 
I'm adding now a system property that allows to set sleep current thread time allowing other processes to execute their own rename action (by default = 50)

PR #882, opened 2020-10-01
Allow the S3 storage to work against Cohesity too. Manually tested against S3, Minio and Cohesity

aaime:


dromagnoli (approved):
Looks good to me.

PR #895, opened 2020-11-20
Minor scalability improvements

aaime:
Doing a "monitor profile" session the following two scalability issues were identified:
* NumberFormat is used to build the bounding box toString representation, which is then used during seed jobs in GeoServer seeds. It has an internal synchronization that shows up in the profile session, better to just reallocate the object instead.
* UUID.newRandomUUID, used to generate temp file names, is both a bit expensive and has synchronizations, worked around by adding a counter as a suffix to a fixed random UUID, which is replaced every N requests instead


dromagnoli:
Looks good to me.
I suppose you wanted to choose me as reviewer instead of assignee for the ticket, right? :)

aaime:
Yep :-p

PR #897, opened 2020-11-25
migrate from travis to github actions

bradh:
As for similar work on geotools and geoserver, this migrates GWC from Travis CI to Github Actions.

The CI functionality should be equivalent.


aaime:
Looking good, thanks a ton for porting this over, and this quickly too!

PR #913, opened 2021-01-03
Switch JUnit tests from v3 to v4

aaime:


aaime:
Cleaned up all xml files formatted by mistake, as per @bradh feedback

geowebcache/core/src/test/java/org/geowebcache/grid/BoundingBoxTest.java
```diff
@@ -41,26 +43,28 @@ public void testBBOXScale() throws Exception {
...
         BoundingBox bb2 = new BoundingBox(5, 5, 20, 20);
 
         BoundingBox intersection = BoundingBox.intersection(bb1, bb2);
-        assertNotNull(intersection);
-        assertEquals(5D, intersection.getWidth());
-        assertEquals(5D, intersection.getHeight());
-        assertTrue(intersection.isSane());
+        Assert.assertNotNull(intersection);
+        Assert.assertEquals(5D, intersection.getWidth(), 0d);
+        Assert.assertEquals(5D, intersection.getHeight(), 0d);
```
jodygarnett, line 54: Did a double take at `5D` vs `0d`, but it is not case sensitive ... I just have never see `D` before.

geowebcache/core/src/test/java/org/geowebcache/grid/GridCalculatorTest.java
```diff
@@ -4,24 +4,25 @@
...
 
     GridSetBroker gridSetBroker =
             new GridSetBroker(Collections.singletonList(new DefaultGridsets(false, false)));
 
-    @Override
-    protected void setUp() throws Exception {
-        super.setUp();
-    }
+    @Before
+    public void setUp() throws Exception {}
```
jodygarnett, line 23: Should this just be removed?

jodygarnett (approved):

geowebcache/core/src/test/java/org/geowebcache/grid/GridCalculatorTest.java
aaime: Could indeed.

geowebcache/core/src/test/java/org/geowebcache/grid/BoundingBoxTest.java
aaime: Both are valid, I prefer the "d" form. The capital one was there already I believe.

Issue #916, opened 2021-01-19
Add ENV parametrization support for s3 endpoint configuration param


mbarto:
Currently only some of the S3BlobStore configuration parameters can use an env parameter. Endpoint is not included. It would be useful to have it, so that it is possible to switch the endpoint on different environments.

PR #917, opened 2021-01-19
#916: added env parameter support to S3BlobStore endpoint configurati…

mbarto:
…on parameter


aaime:
Looks good to me, merging.

PR #924, opened 2021-02-16
Introduce 512px tiles as a possible alternative for the default gridsets

aaime:
As hires screens and 512px tiles are getting common also outside of the "Apple kingdom", it's nice to have them as a default option for GWC and, as as consequence, for GeoServer as well.


smithkm (approved):

aaime:
Oh, legit downstream build failure. Will have to adjust some tests in the gs-gwc module.

PR #930, opened 2021-02-26
Test dependency upgrade: Mockito 1.8.5 -> 3.8.0

groldan:
Upgrade Mockito `1.8.5` (May, 2010) -> `3.8.0` (Feb, 2021) and centralize dependency version at root pom, since `3.2.4` was used in `swiftblob/pom.xml`, but `1.8.5` in all the rest


aaime:
Looking good, thanks!

PR #942, opened 2021-03-29
Ticket-935-mbtiles: Supporting MBTiles Layer

dromagnoli:


geowebcache/mbtiles/src/main/java/org/geowebcache/mbtiles/layer/MBTilesCacheInfo.java
```diff
@@ -0,0 +1,123 @@
...
+import org.geotools.mbtiles.MBTilesMetadata;
+import org.geotools.referencing.CRS;
+import org.geowebcache.grid.BoundingBox;
+import org.opengis.geometry.Envelope;
+import org.opengis.referencing.FactoryException;
+import org.opengis.referencing.crs.CoordinateReferenceSystem;
+import org.opengis.referencing.operation.TransformException;
+
+/** Info Object storing basic MBTiles Cached info */
+public class MBTilesCacheInfo {
```
aaime, line 34:
```suggestion
public class MBTilesInfo {
```

Rationale: this class collects metadata from the MBTiles file. Whether it's cached or not, it's a  matter of how it's used, it could be re-instantiated at every need too. 

If the instance variable in MBTilesLayer was a cache the field name holding it could be called that way instead.

geowebcache/mbtiles/src/main/java/org/geowebcache/mbtiles/layer/MBTilesLayer.java
```diff
@@ -0,0 +1,349 @@
...
+        super.formats = loadMimeTypes();
+        return true;
+    }
+
+    private List<MimeType> loadMimeTypes() {
+        MBTilesMetadata.t_format metadataFormat = cacheInfo.getFormat();
+        try {
+            switch (metadataFormat) {
+                case PNG:
+                    mimeType = MimeType.createFromFormat("image/png");
```
aaime, line 175:
```suggestion
                    mimeType = ImageMime.png;
```

geowebcache/mbtiles/src/main/java/org/geowebcache/mbtiles/layer/MBTilesLayer.java
```diff
@@ -0,0 +1,349 @@
...
+                    break;
+                case JPEG:
+                case JPG:
+                    mimeType = MimeType.createFromFormat("image/jpeg");
```
aaime, line 179:
```suggestion
                    mimeType = ImageMime.jpeg;
```

geowebcache/mbtiles/src/main/java/org/geowebcache/mbtiles/layer/MBTilesLayer.java
```diff
@@ -0,0 +1,349 @@
...
+                    break;
+                case PBF:
+                    mimeType = ApplicationMime.mapboxVector;
+                    vectorTiles = true;
+                    break;
+            }
+        } catch (MimeException e) {
+            throw new RuntimeException(e);
+        }
+        if (mimeType != null) {
```
aaime, line 189: Is there any way mimeType can be null  in the above code? Does not seem possible.

geowebcache/mbtiles/src/main/java/org/geowebcache/mbtiles/layer/MBTilesLayer.java
```diff
@@ -0,0 +1,349 @@
...
+            throw new OutsideCoverageException(tile.getTileIndex(), 0, 0);
+        }
+
+        saveExpirationInformation((int) (tile.getExpiresHeader() / 1000));
+
+        return tile;
+    }
+
+    private byte[] getPbfFromTile(byte[] raw) throws IOException {
+        if (!UNZIP_CONTENT) {
```
aaime, line 255: Not sure how well this plays againt the rest of GWC though, to make it zip preservation work, we'd have to tell the services that the payload is already gzip compressed, and then, decide what to do based on whether the HTTP client declares to support gzip compression or not.

geowebcache/mbtiles/src/main/java/org/geowebcache/mbtiles/layer/MBTilesLayer.java
```diff
@@ -0,0 +1,349 @@
...
+    }
+
+    protected void saveExpirationInformation(int backendExpire) {
+        this.saveExpirationHeaders = false;
+
+        try {
+            if (getExpireCache(0) == GWCVars.CACHE_USE_WMS_BACKEND_VALUE) {
+                if (backendExpire == -1) {
+                    this.expireCacheList.set(0, new ExpirationRule(0, 7200));
+                    log.error(
```
aaime, line 279: Logging at error level seems excessive here, this will be logged for every tile request. Guard and log at a lower level IMHO. Same for the log below.

aaime: Also guard the trace logs, as this is a method called for each tile.

geowebcache/mbtiles/src/test/java/org/geowebcache/mbtiles/layer/MBTilesLayerTest.java
```diff
@@ -0,0 +1,138 @@
...
+    private TileLayerConfiguration getConfig(MockWepAppContextRule extensions) throws Exception {
+        extensions.addBean(
+                "MBTilesLayerConfigProvider",
+                new MBTilesLayerXMLConfigurationProvider(),
+                XMLConfigurationProvider.class);
+        return configSource.create(extensions, configDir);
+    }
+
+    @Test
+    public void testLayerDefaultValues() {
```
aaime, line 112: I would add a test that ensures the right tile is picked when fetching a given x/y/z, to avoid y axis orientation issues in the future. Having the mbtiles dependency means you should have the ability to decode a vector tile and look at its contents.

aaime (changes requested):

geowebcache/mbtiles/src/main/java/org/geowebcache/mbtiles/layer/MBTilesCacheInfo.java
dromagnoli: The field was named cacheInfo.
I named that class like that to avoid "confusion" with the MbtilesInfo of sqllite.
I will rename it.

geowebcache/mbtiles/src/main/java/org/geowebcache/mbtiles/layer/MBTilesLayer.java
dromagnoli: Sounds good. Will fix it (I took that piece of code from ArcGISCacheLayer :-) )

PR #943, opened 2021-03-29
Ticket 941 tilejson

dromagnoli:
TileJSON document support.
Note that classes TileJSON and VectorLayerMetadata come from geoserver ogc-api community modules

### DIFF
--- a/geowebcache/core/src/main/java/org/geowebcache/layer/TileJSONProvider.java
+++ b/geowebcache/core/src/main/java/org/geowebcache/layer/TileJSONProvider.java
@@ -1,0 +1,27 @@
+/**
+ * This program is free software: you can redistribute it and/or modify it under the terms of the
+ * GNU Lesser General Public License as published by the Free Software Foundation, either version 3
+ * of the License, or (at your option) any later version.
+ *
+ * <p>This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * <p>You should have received a copy of the GNU Lesser General Public License along with this
+ * program. If not, see <http://www.gnu.org/licenses/>.
+ *
+ * <p>Copyright 2021
+ */
+package org.geowebcache.layer;
+
+import org.geowebcache.layer.meta.TileJSON;
+
+/** Interface to be implemented by layers which can provide a TileJSON document */
+public interface TileJSONProvider {
+
+    /* Returns true if TileJSON is supported or not compatible */
+    boolean supportsTileJSON();
+
+    /* Returns a TileJSON for this layer, or null if not supported or not compatible */
+    TileJSON getTileJSON();
+}
--- a/geowebcache/mbtiles/src/main/java/org/geowebcache/mbtiles/layer/MBTilesLayer.java
+++ b/geowebcache/mbtiles/src/main/java/org/geowebcache/mbtiles/layer/MBTilesLayer.java
@@ -44,6 +44,8 @@ import org.geowebcache.grid.OutsideCoverageException;
 import org.geowebcache.io.ByteArrayResource;
 import org.geowebcache.layer.AbstractTileLayer;
 import org.geowebcache.layer.ExpirationRule;
+import org.geowebcache.layer.TileJSONProvider;
+import org.geowebcache.layer.meta.TileJSON;
 import org.geowebcache.mime.ApplicationMime;
 import org.geowebcache.mime.ImageMime;
 import org.geowebcache.mime.MimeType;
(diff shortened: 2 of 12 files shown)


geowebcache/mbtiles/src/main/java/org/geowebcache/mbtiles/layer/MBTilesLayer.java
```diff
@@ -352,4 +354,21 @@ public void setExpirationHeader(HttpServletResponse response, int zoomLevel) {
...
     }
+
+    @Override
+    public boolean supportsTileJSON() {
+        return true;
+    }
+
+    @Override
+    public TileJSON getTileJSON() {
+
```
aaime, line 365: Why an empty line here? :D

geowebcache/mbtiles/src/test/java/org/geowebcache/mbtiles/layer/MBTilesJsonTest.java
```diff
@@ -0,0 +1,69 @@
...
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import org.geowebcache.layer.meta.TileJSON;
+import org.geowebcache.layer.meta.VectorLayerMetadata;
+import org.junit.Test;
+
+public class MBTilesJsonTest extends MBTilesLayerTest {
```
aaime, line 30: This will make Maven re-run all the tests in the base class as well (so, they will run twice). Any reason not to add this test directly to the base class?

geowebcache/wmts/src/main/java/org/geowebcache/service/wmts/WMTSService.java
```diff
@@ -557,6 +583,15 @@ public void handleRequest(Conveyor conv) throws OWSException, GeoWebCacheExcepti
                 ConveyorTile convTile = (ConveyorTile) conv;
                 WMTSGetFeatureInfo wmsGFI = new WMTSGetFeatureInfo(convTile);
                 wmsGFI.writeResponse(stats);
+            } else if (tile.getHint().equals(GET_TILEJSON)) {
+                getSecurityDispatcher().checkSecurity(tile);
+                ConveyorTile convTile = (ConveyorTile) conv;
+                TileLayer layer = convTile.getLayer();
+                if (layer instanceof TileJSONProvider) {
```
aaime, line 590: Should return a 404 if someone asks for a tilejson when the url is not included in the capabilities

geowebcache/core/src/main/java/org/geowebcache/layer/TileJSONProvider.java
```diff
@@ -0,0 +1,27 @@
...
+ */
+package org.geowebcache.layer;
+
+import org.geowebcache.layer.meta.TileJSON;
+
+/** Interface to be implemented by layers which can provide a TileJSON document */
+public interface TileJSONProvider {
+
+    /* Returns true if TileJSON is supported or not compatible */
+    boolean supportsTileJSON();
```
aaime, line 23: This method is useful only if it guarantees that getTileJSON returns a non null value, and it's at the same time faster than doing ``getTileJSON() != null``, otherwise I'd consider it redundant.

geowebcache/wmts/src/main/java/org/geowebcache/service/wmts/WMTSTileJSON.java
```diff
@@ -0,0 +1,117 @@
...
+        TileJSON json = provider.getTileJSON();
+        // "/{layer}/{tileMatrixSet}/{tileZoom}/{tileRow}/{tileCol}"
+        List<String> urls = new ArrayList<>();
+        for (MimeType mimeType : layer.getMimeTypes()) {
+            Set<String> gridSubSets = layer.getGridSubsets();
+            gridSubSets
+                    .stream()
+                    .forEach(
+                            s -> {
+                                GridSubset grid = layer.getGridSubset(s);
```
aaime, line 63: That's one big lambda, is it possible to turn into into a separate method?

aaime (changes requested):
The WMTS capabilities has been modified, could you add a test for it?

geowebcache/mbtiles/src/main/java/org/geowebcache/mbtiles/layer/MBTilesLayer.java
dromagnoli: No idea (Didn't notice that) maybe hitting Enter on a different windows whilst the focus was actually on the IDE :D 
Removing it.

geowebcache/core/src/main/java/org/geowebcache/layer/TileJSONProvider.java
dromagnoli: In the MBTilesLayer implementation it always return true.
On a GeoServerTileLayer implementation I will return true if the underlying GridSet is 3857 or 900913.

Given a Layer the GetCapabilities will check if that layer supportsTileJSON and quickly skip it if not.

Would that make sense?

aaime: Yessir!

aaime (approved):

aaime:
Thanks, looks good :-D

PR #955, opened 2021-04-12
TileJSON url: Supporting format parameter

dromagnoli:


geowebcache/wmts/src/main/java/org/geowebcache/service/wmts/WMTSGetCapabilities.java
```diff
@@ -750,10 +750,11 @@ private void layerResourceUrls(
         }
         if (layer instanceof TileJSONProvider) {
             TileJSONProvider provider = (TileJSONProvider) layer;
-            String format = "json";
             if (provider.supportsTileJSON()) {
-                String template = baseTemplate + "/{style}/tilejson?format=" + format;
-                layerResourceUrlsGen(xml, format, "TileJSON", template);
+                for (String format : mimeFormats) {
+                    String template = baseTemplate + "/{style}/tilejson?format=" + format;
```
aaime, line 755: Hmm.. I would have expected to see the format in the URL. The &format parmaeter normally controls the format of what is returned (tilejson is always json).

geowebcache/wmts/src/test/java/org/geowebcache/service/wmts/WMTSRestTest.java
```diff
@@ -192,13 +197,15 @@ public void testGetTileJSONWithoutStyle() throws Exception {
         // Checking the response contains a tileUrl without the style
         assertTrue(
                 content.contains(
-                        "\"tiles\":[\"http://localhost/service/wmts/rest/mockLayerTileJSON/EPSG:900913/EPSG:900913:{z}/{y}/{x}?format=image/png\"]"));
+                        "\"tiles\":[\"http://localhost/service/wmts/rest/mockLayerTileJSON/EPSG:900913/EPSG:900913:{z}/{y}/{x}?format="
```
aaime, line 200: Guess it cannot be done in GWC, but downstreams in GeoServer, it would be useful to compare the tile URLs and make sure they end up being different depending on what format parameter has been provided asking for the TileJSON, e.g., point to mvt if the tilejson was requested for mvt tiles, and to pngs if the tilejson was requested for png tiles (on a layer that supports both formats)

dromagnoli: I managed to have it done in a GWC Test. Updating it.

geowebcache/wmts/src/main/java/org/geowebcache/service/wmts/WMTSGetCapabilities.java
dromagnoli: Fixed

aaime:
Merged, thanks!

PR #961, opened 2021-04-26
Use LevelID as ArcGIS cache gridset's name

initialdmg:
Arcgis cache layer extension does not assign any scale name to gridsets, so `GridSetFactory` make it as `gridsetName:i` , like `EPSG:4326_arcgis:0`. This patch use `levelID`, like `0`, `1`, `2` ..., as scaleName, and it is correspond with our practice in WMTS.

### DIFF
--- a/geowebcache/arcgiscache/src/main/java/org/geowebcache/arcgis/layer/GridSetBuilder.java
+++ b/geowebcache/arcgiscache/src/main/java/org/geowebcache/arcgis/layer/GridSetBuilder.java
@@ -55,7 +55,7 @@ class GridSetBuilder {
          */
         final double[] scaleDenominators = null;
         final Double metersPerUnit;
-        final String[] scaleNames = null;
+        final String[] scaleNames;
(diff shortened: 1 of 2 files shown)


aaime:
I agree many tools struggle with gridsets names that are not numeric. It is unfortunate, as the WMTS the specification poses no such constraint, but best to align with what works in practice.

However, this change should have a unit or integration test covering the change, please add one.

initialdmg:
Code committed earlier was an idea, and was not edited in editor, sry for this behaviour. To keep timeline clear, i force pushed the right code and test case, which has passed on my machine. You can review the code now 😃

smithkm (approved):

jodygarnett:
Has this PR become stale? Request a rebase as some time has passed.

jodygarnett:
This is the only outstanding PR that would make a new 1.22.x interesting; but the PR appears stale?

aaime:
@smithkm what should we do regarding this PR? It's old but seems mergeable. Pheraps it should be rebased to make sure  the builds are all still working fine.

PR #972, opened 2021-06-21
Fix ServiceMetadataURL in cases where baseUrl contains query string parameters

sampov2:
Geowebcache may be used in situations where the baseUrl contains query string parameters. One such case is in GeoServer when using the authkey community extension. This fix appends a `?` only when no `?` appears in the baseUrl and an `&` in other cases.

### DIFF
--- a/geowebcache/wmts/src/main/java/org/geowebcache/service/wmts/WMTSGetCapabilities.java
+++ b/geowebcache/wmts/src/main/java/org/geowebcache/service/wmts/WMTSGetCapabilities.java
@@ -166,10 +166,16 @@ public class WMTSGetCapabilities {
             operationsMetadata(xml);
 
             contents(xml);
+            String kvpBaseUrl = baseUrl;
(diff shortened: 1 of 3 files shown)


smithkm:
This should have unit tests and also it seems to neglect the situation that the base URL has exactly one parameter and therefore has an ampersand but no question mark.

sampov2:
Thanks for the comments @smithkm . I added tests as you proposed. However I'm not sure what you meant with your example about an URL with exactly one parameter. If there are any (one or more) query string parameter, there is always a question mark as that is what separates the path and query string. Can you give an example of what you mean?

smithkm:
Sorry, yes I think i must have been transposing ? and & in my mind when I wrote that.  Never mind.  Unit tests look good, thanks.

smithkm (approved):

smithkm:
Looks like the QA check is failing due to a formatting problem.  Do a `mvn clean install` and commit the changes that result.

Issue #977, opened 2021-08-05
Update to use Apache HttpComponents through out and bump version


ianturton:
`commons-httpclient` is not longer supported so it would be a good idea to update to use `org.apache.httpcomponents` and while we're doing that bump the version of that to 4.5.13 (to avoid some know issues).

PR #978, opened 2021-08-05
move from commons-httpclient to httpcomponents

ianturton:
Implement issue #977


geowebcache/core/src/main/java/org/geowebcache/util/HttpClientBuilder.java
```diff
@@ -36,17 +33,17 @@
...
 
-    private URL proxyUrl = null;
-
     private Integer backendTimeoutMillis = null;
+    private static final HttpClientConnectionManager connectionManager =
+            new PoolingHttpClientConnectionManager();
 
     private boolean doAuthentication = false;
 
-    private int concurrency;
```
aaime, line 49: Hum... was this configurable before? Yep, it's a constructor argument, still there but I don't see it being used. Think it can be set in a "connection manager" class (or something like that, vague memories)

ianturton: good catch, I knew there was something I needed to come back to

PR #980, opened 2021-08-20
Adding OGC TileMatrixSet common gridsets

aaime:
Also makes sure WMTS does not return all grids. Checked TMS only returns
the ones used already, and so does WMS-C.

Downstream builds fails to to new gridsets, gets fixed in https://github.com/geoserver/geoserver/pull/5242

### DIFF
--- a/geowebcache/core/src/main/java/org/geowebcache/config/DefaultGridsets.java
+++ b/geowebcache/core/src/main/java/org/geowebcache/config/DefaultGridsets.java
@@ -40,6 +40,12 @@ public class DefaultGridsets extends SimpleGridSetConfiguration {
 
     private final GridSet WORLD_EPSG3857x2;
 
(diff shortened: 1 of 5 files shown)


mprins:
@aaime haven't looked in detail, but I think these changes are causing the "downstream geoserver" build of geotools to fail with:

```
2021-09-04T15:54:37.8613378Z [INFO] 
2021-09-04T15:54:37.8614225Z [ERROR] Failures: 
2021-09-04T15:54:37.8615682Z [ERROR]   GWCSettingsPageTest.testNewDefaultGridSet:334 expected:<[EPSG:4326x2]> but was:<[CanadianNAD83_LCC]>
```

maybe an ordering issue in the test

mprins:
thanks for https://github.com/geoserver/geoserver/pull/5242

PR #989, opened 2021-11-20
Remove duplicate env variable lookup code, always use GWCVars

groldan:
Variable lookups using system-property/servlet-context/system-environment
precedence was being performed at several places with slightly different
logic (checking for nulls or not).

Use GWCVars as the single utility to perform variable lookups.

Deprecate `GeoWebCacheExtensions.getProperty(String, ServletContext)`  in favor of `GWCVars.findEnvVar(ApplicationContext, String)`.

### DIFF
--- a/geowebcache/core/src/main/java/org/geowebcache/GeoWebCacheExtensions.java
+++ b/geowebcache/core/src/main/java/org/geowebcache/GeoWebCacheExtensions.java
@@ -24,6 +24,7 @@ import javax.servlet.ServletContext;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 import org.geowebcache.config.BaseConfiguration;
+import org.geowebcache.util.GWCVars;
 import org.geowebcache.util.SuppressFBWarnings;
 import org.springframework.beans.BeansException;
 import org.springframework.context.ApplicationContext;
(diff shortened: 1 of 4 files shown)


geowebcache/core/src/main/java/org/geowebcache/GeoWebCacheExtensions.java
```diff
@@ -349,66 +349,6 @@ public static String getProperty(String propertyName) {
...
-     *   <li>Environment variable
-     * </ul>
-     *
-     * and returns the first non null, non empty value found.
-     *
-     * @param propertyName The property name to be searched
-     * @param context The servlet context used to look into web.xml (may be null)
-     * @return The property value, or null if not found
-     */
-    public static String getProperty(String propertyName, ServletContext context) {
```
groldan, line 374: Note this one is safe to remove as it was only called from a test case. Not used in geoserver either.

groldan:
@smithkm / @aaime can I ask for a rewiew?

geowebcache/core/src/main/java/org/geowebcache/GeoWebCacheExtensions.java
smithkm: I don't want to rule out the possibility that it's being used in some custom extension or modification somewhere so I'd be happier if it were deprecated for a release or two first before removing it entirely

groldan: You got it https://github.com/GeoWebCache/geowebcache/pull/989/files#diff-88bfaa979935d2ccfabe9c9cf93d6628892c241905bc012c94192588d6f7621fR371-R374

smithkm (approved):

PR #992, opened 2021-11-30
Fix bad operation check in DELETE_LAYER from Memory Blobstore

lievenjacquot:
The object length check in DELETE_LAYER operation from MemoryBlobStore has to be set to 1 instead of 2 as the object will be always the layername


aaime:
Merged, thanks!

PR #998, opened 2021-12-17
Improve disk quota startup time on caches with thousands of parameter directories

aaime:
When trying to create a disk quota database anew, or starting up with a database out of synch with the file system, the disk quota tries to collect stats from disk.
It mostly works using background tasks, but first it has to list all available tile caches for each layer (one for each combo of request parameters, if any). In case there are many of them (e.g., long time series, cached) this can take a lot of time, breaking the promise that ``LayerCacheInfoBuilder.buildCacheInfo`` "returns immediately". We have actually observed the quota startup stuck for ages, to the point that it looked stuck, while it was actually listing directories (many layers, all time enabled, many times).

This change moves the directory collection part to the background, actually allowing the ``buildCacheInfo`` method to return immediately.

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -36,12 +36,11 @@ jobs:
         git clone https://github.com/geoserver/geoserver.git geoserver
     - name: Build GeoServer with tests
       run: |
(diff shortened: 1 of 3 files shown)


dromagnoli (approved):
LGTM

PR #1007, opened 2022-01-13
Improvements, async delete, upload interrupt support for Swift blobstores

mtnorthcott:
This PR introduces a number of fixes/improvements, asynchronous tile deletion, and upload interrupt support for Swift blobstores.

Change summary:
* jclouds update to 2.3.0
* Adapt existing upload queue to serve as a more generic task queue
* Queue tile deletions as tasks to the task queue when tile caching is disabled for a layer
* Interrupt and cancel any existing upload tasks when tile caching is disabled for the associated layer
* General code improvements
* Additional tests to cover changes

### DIFF
--- a/geowebcache/swiftblob/src/main/java/org/geowebcache/swift/SwiftBlobStore.java
+++ b/geowebcache/swiftblob/src/main/java/org/geowebcache/swift/SwiftBlobStore.java
@@ -63,53 +63,59 @@ public class SwiftBlobStore implements BlobStore {
 
     private final BlobStoreListenerList listeners = new BlobStoreListenerList();
 
-    private final String containerName;
+    private final SwiftBlobStoreInfo config;
 
     private final TMSKeyBuilder keyBuilder;
 
     private volatile boolean shutDown;
 
     /** JClouds Swift API */
-    private final SwiftApi swiftApi;
+    private SwiftApi swiftApi;
 
     /** Swift Object API */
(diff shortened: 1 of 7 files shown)


geowebcache/swiftblob/src/main/java/org/geowebcache/swift/SwiftBlobStore.java
```diff
@@ -183,14 +190,8 @@ public TileToKey(String coordsPrefix, MimeType mimeType) {
...
-            long z = loc[2];
-            long x = loc[0];
-            long y = loc[1];
-            StringBuilder sb = new StringBuilder(coordsPrefix);
-
-            // builds the path to the tile
-            sb.append(z).append('/').append(x).append('/').append(y).append('.').append(extension);
-            return sb.toString();
+            // String in format "<prefix><z>/<x>/<y>.<ext>"
+            return String.format("%s%d/%d/%d.%s", coordsPrefix, loc[2], loc[0], loc[1], extension);
```
aaime, line 194: This is [2 orders of magnitude slower than the previous code](https://stackoverflow.com/questions/44117788/performance-between-string-format-and-stringbuilder), but in the context of fetching a tile, the slowdown is most likely irrelevant.

aaime:
Don't pretend to understand the changes, but the builds are green and nothing in the changes stood out as problematic. Merging.

PR #1008, opened 2022-01-15
Remove stax:stax and stax:stax-api dependencies, they're part of the default Java library

groldan:
The StAX API is part of the JRE default library since Java 1.6.


aaime:
All builds are green, including the downstreams one. Seemed like an un-needed dependency (probably was added back when the JDK did not support StaX?). Merging.

groldan:
Indeed it comes from pre Java 1.6 days. Thanks for the review and merge.

PR #1015, opened 2022-02-09
Correct handling of layer names in SeedController

groldan:
Fixes #1013

PR #1021, opened 2022-02-16
Disable test SwiftBlobStoreTest using Thread.sleep, they are unreliable on build servers/github actions

aaime:


groldan:
re-run the failing one cause it network glitch resolving a maven dependency.
I say we merge this one and I rebase #1015 and merge if all goes well. @aaime agree?

aaime:
Go go go!!

PR #1024, opened 2022-03-09
Update ImageIO-Ext to 1.3-SNAPSHOT to evaluate gdal32 compatibility

jodygarnett:
This pull request will need ot be changed to the official imageio-ext 1.3.12 when it is available.

See https://github.com/geosolutions-it/imageio-ext/pull/253 for details


jodygarnett:
Adding geosolutions repository in as a snapshot repository; should comment this out when PR is updated to 1.3.12 (or 1.4.0).

jodygarnett:
Having some trouble with sortpom on linux8; trying again.

aaime:
Odd, just running the buidl from the command line should do the trick, it auto-formats while building (even without -Dqa).

PR #1025, opened 2022-03-21
Change from commons-logging to java util logging frontend (wiith log4j backend)

jodygarnett:
While this pull request is a good idea the primary focus is checking out integration with GeoTools upgrade to log4j 2 (see https://github.com/geotools/geotools/pull/3809).

Tasks:

* [x] commons-logging upgrade
* [x] change to reload4j backend
* [x] change to java util logging front-end using geotools Logging class
* [x] change to log4j backend
* [x] Setup LoggingContextListener with properties to control logging configuration  

Note GeoWebCache still requires commons-logging dependency for the spring-framework core, but no longer use this framework directly.

Related:

* https://github.com/geotools/geotools/pull/3809 
* https://github.com/geoserver/geoserver/pull/5734


jodygarnett:
While this may build prior to https://github.com/geotools/geotools/pull/3809 being merged the geotools logs will be directed to java util logging (and thus not represent a good test).

jodygarnett:
A number of dependencies (azure, s3) have forced the inclusion of slf4j bridge as well.

geowebcache/core/src/main/resources/org/geowebcache/rest/webresources/openlayers/OpenLayers.js
```diff
@@ -128,19 +128,19 @@ return camelizedString;},format:function(template,context,args){if(!context){con
...
-if(!String.prototype.trim){String.prototype.trim=function(){OpenLayers.Console.warn(OpenLayers.i18n("methodDeprecated",{'newMethod':'OpenLayers.String.trim'}));return OpenLayers.String.trim(this);};}
-if(!String.prototype.camelize){String.prototype.camelize=function(){OpenLayers.Console.warn(OpenLayers.i18n("methodDeprecated",{'newMethod':'OpenLayers.String.camelize'}));return OpenLayers.String.camelize(this);};}
+if(typeof replacement=='undefined'){return'undefined';}else{return replacement;}};return template.replace(OpenLayers.String.tokenRegEx,replacer);},tokenRegEx:/\$\{([\w.]+?)\}/g,numberRegEx:/^([+-]?)(?=\d|\.\d)\d*(\.\d*)?([Ee]([+-]?\d+))?$/,isNumeric:function(value){return OpenLayers.String.numberRegEx.test(value);},numericIf:function(value){return OpenLayers.String.isNumeric(value)?parseFloat(value):value;}};if(!String.prototype.startsWith){String.prototype.startsWith=function(sStart){OpenLayers.Console.warning(OpenLayers.i18n("methodDeprecated",{'newMethod':'OpenLayers.String.startsWith'}));return OpenLayers.String.startsWith(this,sStart);};}
```
aaime, line 131: Looks like a search/replace gone wrong.

geowebcache/core/src/main/java/org/geowebcache/util/Logging.java
```diff
@@ -0,0 +1,120 @@
...
+import org.apache.logging.log4j.core.config.DefaultConfiguration;
+import org.apache.logging.log4j.core.config.NullConfiguration;
+import org.apache.logging.log4j.core.config.builder.impl.BuiltConfiguration;
+import org.apache.logging.log4j.core.config.json.JsonConfiguration;
+import org.apache.logging.log4j.core.config.properties.PropertiesConfiguration;
+import org.apache.logging.log4j.core.config.xml.XmlConfiguration;
+import org.apache.logging.log4j.core.config.yaml.YamlConfiguration;
+
+/**
+ * Uses {@link org.geotools.util.logging.Logging} to delegate to provided logging environment.
```
aaime, line 31: I don't understand why this class is needed... all you need is a web application initializer setting the preferred logging framework (you can contorl the order of the initializers execution in web.xml) and would have avoided having to setup logs in a way that's different from GeoTools.
Also, this setup is quite error-prone, when writing a new class people used to GeoTools will likely just use GeoTools own Logging instead.

I'd strongly recommend to back out this change, if not all, at least having to go through gwc Logging rather than Geotools logging.

documentation/en/user/source/troubleshooting/index.rst
```diff
@@ -29,7 +29,19 @@ Note that you can safely modify gridsubsets, but the gridset should always be th
 Logging
 -------
 
-Logging can be controlled through :file:`WEB-INF/classes/log4j.properties`. By default, log messages will end up in the standard log of the container, which is for Apache Tomcat is inside the :file:`logs` directory.
```
aaime, line 32: On this, I'll not that the code still has ``log4j.properties`` files lying around, including one in the final war file:

```
find . -name "log4j.properties"
./geowebcache/web/src/main/resources/log4j.properties
./geowebcache/web/target/geowebcache/WEB-INF/classes/log4j.properties
./geowebcache/web/target/classes/log4j.properties
./geowebcache/core/src/test/resources/log4j.properties
./geowebcache/core/target/test-classes/log4j.properties
./geowebcache/sqlite/src/test/resources/log4j.properties
./geowebcache/sqlite/target/test-classes/log4j.properties
```

geowebcache/pom.xml
```diff
@@ -167,13 +167,19 @@
...
-
-      <!-- logging dependencies -->
       <dependency>
-        <groupId>commons-logging</groupId>
-        <artifactId>commons-logging</artifactId>
-        <version>${commons-logging.version}</version>
+        <groupId>org.apache.httpcomponents</groupId>
+        <artifactId>httpclient</artifactId>
+        <version>${httpcomponents.version}</version>
+        <exclusions>
```
aaime, line 174: Same reaction as in GEoTools... is httpclient written against commons-logging? If it is, then we'd see failures at deploy, if the expected API is not in the classpath.

geowebcache/pom.xml
```diff
@@ -561,7 +571,6 @@
       </plugin>
 
       <plugin>
-        <groupId>org.apache.maven.plugins</groupId>
```
aaime, line 541: Interesting, what's up here? I guess this groupid is a default, when talking about maven plugins?

geowebcache/core/pom.xml
```diff
@@ -149,6 +178,12 @@
       <groupId>org.apache.httpcomponents</groupId>
       <artifactId>httpclient</artifactId>
       <version>${httpcomponents.version}</version>
+      <exclusions>
```
aaime, line 181: Same reaction as in GeoTools... is httpclient written against commons-logging? If it is, then we'd see failures at deploy, if the expected API is not in the classpath.

geowebcache/core/pom.xml
```diff
@@ -199,7 +234,6 @@
   <build>
     <plugins>
       <plugin>
-        <groupId>org.apache.maven.plugins</groupId>
```
aaime, line 202: Interesting, what's up here? I guess this groupid is a default, when talking about maven plugins?

geowebcache/web/src/main/resources/log4j2.xml
```diff
@@ -6,30 +6,41 @@
...
-            <PatternLayout pattern="%date{dd MMM HH:mm:ss} (%t) %-6level [%logger{2}] - %msg%n"/>
+            <PatternLayout pattern="%date{dd MMM HH:mm:ss} (%t) %-6level [%logger{2}] - %msg%n%throwable"/>
         </Console>
-        <RollingFileAppender name="Logfile">
-            <MaxBackupIndex>3</MaxBackupIndex>
-            <MaxFileSize>10MB</MaxFileSize>
-            <PatternLayout pattern="%date{dd MMM HH:mm:ss} (%t) %-6level [%logger{2}] - %msg%n"/>
-        </RollingFileAppender>
+        <RollingFile name="Logfile">
+            <FileName>geowebcache.log</FileName>
```
aaime, line 12: Where does this file end up being located?

geowebcache/azureblob/pom.xml
```diff
@@ -23,6 +23,17 @@
...
       <version>11.0.0</version>
+      <exclusions>
+        <exclusion>
+          <groupId>org.slf4j</groupId>
+          <artifactId>slf4j-api</artifactId>
+        </exclusion>
+      </exclusions>
+    </dependency>
+
+    <dependency>
```
aaime, line 34: Hum hum... but this in turn depends on the slf4j-api no?
And the dependency management in the root pom should make sure they use the same version.

geowebcache/azureblob/pom.xml
```diff
@@ -25,8 +25,8 @@
       <version>11.0.0</version>
       <exclusions>
         <exclusion>
-          <groupId>org.slf4j</groupId>
           <artifactId>slf4j-api</artifactId>
+          <groupId>org.slf4j</groupId>
```
aaime, line 29: Seeing a group id after the artifact id seems strange. Is sortpom not complaining?

aaime (changes requested):
Various little bits, two major ones:

- Search/replace gone wrong in Openlayers.js
- Having to use a Logging class from a different package to lookup the logger is a big no-go, especially when both are in the classpath.. uncomprehensible and unmaintainable.

geowebcache/azureblob/pom.xml
jodygarnett: it probably did, I am having trouble running these things locally.

geowebcache/core/pom.xml
jodygarnett: I was working hard to get only one commons-logging in the mix, and using the one provided by `log4j-jcl`.

geowebcache/core/src/main/resources/org/geowebcache/rest/webresources/openlayers/OpenLayers.js
jodygarnett: thanks I did not catch this

geowebcache/web/src/main/resources/log4j2.xml
aaime: Looks like there is no file name in the existing setup (it was there, has been commented out):

```
# Debugging log settings
log4j.rootLogger=info, stdout
# ------------- stdout logging ---------------------
log4j.appender.stdout=org.apache.log4j.ConsoleAppender
log4j.appender.stdout.layout=org.apache.log4j.PatternLayout
# Follow GeoServer's logging pattern
log4j.appender.stdout.layout.ConversionPattern=%d{dd MMM HH:mm:ss} (%t) %p [%c{2}] - %m%n

# ------------- file logging ---------------------
log4j.appender.logfile=org.apache.log4j.RollingFileAppender

# Keep three backup files.
log4j.appender.logfile.MaxBackupIndex=3

# Maximum file size is 10 MB
log4j.appender.logfile.MaxFileSize=10MB

# TODO : make platform-independent logging
#log4j.appender.logfile.File=${catalina.home}/webapps/logs/geowebcache.log
#log4j.appender.logfile.File=/tmp/geowebcache.log

log4j.appender.logfile.layout=org.apache.log4j.PatternLayout

log4j.appender.logfile.layout.ConversionPattern=%d{dd MMM HH:mm:ss} (%t) %p [%c{2}] - %m%n

log4j.category.org.geowebcache.seed=info
log4j.category.org.geowebcache.diskquota=info
```

The above is ``/geowebcache/web/target/geowebcache/WEB-INF/classes/log4j.properties``

documentation/en/user/source/troubleshooting/index.rst
aaime: I see just now they all are in "target"... never mind!

jodygarnett: I will replace

geowebcache/pom.xml
jodygarnett: yes it is

geowebcache/core/src/main/java/org/geowebcache/util/Logging.java
aaime: Does it happen too late in GeoServer? Has worked fine for years, happens in LoggingStartupContextListener

geowebcache/web/src/main/resources/log4j2.xml
jodygarnett: I did download a geowebcache.war from source forge and deploy in tomcat, creates in in the catalina.home (not very helpful)

aaime: Maybe put it in the GWC data dir instead? However, this means reconfiguring the logging on the fly, once you have determined where the data dir is. (and keep only the console logger going up to that point).

geowebcache/core/src/main/java/org/geowebcache/util/Logging.java
jodygarnett: Andrea the solid reason for the new Logging class is to respect GeoWebCache convention of using the full class name, GeoTools uses just the package name.

aaime: That's just aesthetics... how the log messages end up showing in the log files should be up to a user-controllable configuration.

geowebcache/web/src/main/resources/log4j2.xml
jodygarnett: I turned off the file logigng, matching previous functionality (that only logged if you provided a file name.
I did include an example file, that can be switched to using a system property.

geowebcache/core/src/main/resources/org/geowebcache/rest/webresources/openlayers/OpenLayers.js
jodygarnett: Restore OpenLayers.js

geowebcache/core/src/main/java/org/geowebcache/util/Logging.java
jodygarnett: This class is removed, LoggingContextListener introduced providing geoserver like control with system properties.

Indeed ``org.geowebcache.util.logging.policy`` policy of ``Provided`` avoid any logging configuration; intended for when geowebcache is used by geoserver as a library.

documentation/en/user/source/troubleshooting/index.rst
jodygarnett: The individual modules have been configured to use `src/test/resources/logging.properties` during testing, only `web` uses a `log4j2.xml` for testing.

It was just much easier than trying to figure out how to get `GeoTools.init()` to be called by the various test cases.

jodygarnett: Marking this resolved as the docs have been updated with working examples.

geowebcache/azureblob/pom.xml
jodygarnett: I was not aware of ``<dependencyManagement>`` ability to override transitive version numbers.
I will revisit these exclusions with a transitive dependency in place.

geowebcache/azureblob/pom.xml
jodygarnett: Okay you are correct dependency management overrides the transitive dependency version. Kind of scary, but okay...

jodygarnett:
With https://github.com/geotools/geotools/pull/3809 I am expecting this to now be able to compile.

PR #1054, opened 2022-05-18
[GWC-1048] NPE requesting WMS (1.21.x)

turingtestfail:
created tests for httpclient creation

fixed issue with headers

cleanup

check http code before length

cleanup


jodygarnett:
I do not understand "GWC-1048", does not seem to match up with a geowebcache issue (or even a geoserver one?)

PR #1067, opened 2022-06-30
WMS full does not correctly rescale when requested resolution is between tile values

aaime:
The rescale is performed, but the result is thrown away.


dromagnoli (approved):
LGTM

PR #1073, opened 2022-07-17
Review startup warning, info and config logging levels

jodygarnett:
The startup, and especially the shutdown, are very chatty. 

- Take advantage of CONFIG log levels, some configuration still logged as INFO
- Cut back on shutdown messages, many message intended for developers (about destory() and threads) are reduced to fine logging level

### DIFF
--- a/geowebcache/core/src/main/java/org/geowebcache/storage/DefaultStorageFinder.java
+++ b/geowebcache/core/src/main/java/org/geowebcache/storage/DefaultStorageFinder.java
@@ -130,9 +130,9 @@ public class DefaultStorageFinder {
             if (tmpDir != null) {
                 File temp = new File(tmpDir, "geowebcache");
                 logMsg =
-                        "Reverting to java.io.tmpdir "
-                                + this.defaultPrefix
-                                + " for storage. "
+                        "Reverting to java.io.tmpdir '"
+                                + temp.getAbsolutePath()
+                                + "' for storage. "
                                 + "Please set "
                                 + GWC_CACHE_DIR
                                 + ".";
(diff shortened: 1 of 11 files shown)


geowebcache/core/src/main/java/org/geowebcache/storage/DefaultStorageFinder.java
```diff
@@ -130,9 +130,9 @@ private void determineDefaultPrefix() {
             if (tmpDir != null) {
                 File temp = new File(tmpDir, "geowebcache");
                 logMsg =
-                        "Reverting to java.io.tmpdir "
-                                + this.defaultPrefix
-                                + " for storage. "
+                        "Reverting to java.io.tmpdir '"
+                                + temp
+                                + "' for storage. "
```
juanluisrp, line 135: `java.io.File.toString()` can return just a relative path, depending on what has been using to construct the instance. I'd use here `file.getAbsolutePath()` to be sure that the absolute path is printed.

juanluisrp:
Not really related to this PR but I've noticed that the loggin stuff doesn't really follow a uniform style, sometimes doesn't even follow the Java coding style:
* `private static Logger log = Logging.getLogger(MyClass.class.getName());`: should this be called `LOG` (it is a constant)?
* In some places `log(Level level, String msg)` is used and in others `log.info`, `log.fine`.

jodygarnett:
> Not really related to this PR but I've noticed that the loggin stuff doesn't really follow a uniform style

That is correct; this codebase is not as closely managed as GeoServer.

> Sometimes doesn't even follow the Java coding style:
> 
> * `private static Logger log = Logging.getLogger(MyClass.class.getName());`: should this be called `LOG` (it is a constant)?

I agree, to be consistent `LOGGER` should be used and the field marked final.

> * In some places `log(Level level, String msg)` is used and in others `log.info`, `log.fine`.

This is actually by design - when an exception is to be logged `LOGGER.log( Level.FINE, "Message", e)` is the only approach available.

PR #1080, opened 2022-08-15
Fix S3 blobstore geowebcache parameterization issue

shadab-geo:
This PR fixes the S3 Blobstore parameterization issue. 
For example, if we set the bucket name to ${S3_BUCKET} and pass it as an env variable, we get an error suggesting that its not possible to specify bucket name containing '$'

### DIFF
--- a/geowebcache/s3storage/src/main/java/org/geowebcache/s3/S3BlobStoreInfo.java
+++ b/geowebcache/s3storage/src/main/java/org/geowebcache/s3/S3BlobStoreInfo.java
@@ -347,23 +347,22 @@ public class S3BlobStoreInfo extends BlobStoreInfo {
         S3BlobStoreInfo blobStore = SerializationUtils.clone(this);
 
         if (allowEnvParametrization && gwcEnvironment != null) {
-            blobStore.setName((String) gwcEnvironment.resolveValue(getName()));
-            blobStore.setEnabled((Boolean) gwcEnvironment.resolveValue(isEnabled()));
-            blobStore.setDefault((Boolean) gwcEnvironment.resolveValue(isDefault()));
-            blobStore.setAccess((Access) gwcEnvironment.resolveValue(getAccess()));
-            blobStore.setPrefix((String) gwcEnvironment.resolveValue(getPrefix()));
-            blobStore.setUseHTTPS((Boolean) gwcEnvironment.resolveValue(isUseHTTPS()));
-            blobStore.setUseGzip((Boolean) gwcEnvironment.resolveValue(isUseGzip()));
-
+            blobStore.setName(getName());
+            blobStore.setEnabled(isEnabled());
+            blobStore.setDefault(isDefault());
+            blobStore.setAccess(getAccess());
+            blobStore.setPrefix(getPrefix());
+            blobStore.setUseHTTPS(isUseHTTPS());
+            blobStore.setUseGzip(isUseGzip());
+            blobStore.setMaxConnections(getMaxConnections());
+            blobStore.setProxyPort(getProxyPort());
             blobStore.setBucket((String) gwcEnvironment.resolveValue(getBucket()));
             blobStore.setAwsAccessKey((String) gwcEnvironment.resolveValue(getAwsAccessKey()));
             blobStore.setAwsSecretKey((String) gwcEnvironment.resolveValue(getAwsSecretKey()));
-            blobStore.setMaxConnections((Integer) gwcEnvironment.resolveValue(getMaxConnections()));
             blobStore.setProxyDomain((String) gwcEnvironment.resolveValue(getProxyDomain()));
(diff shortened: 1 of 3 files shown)


geowebcache/s3storage/src/main/java/org/geowebcache/s3/S3BlobStoreInfo.java
```diff
@@ -334,7 +337,38 @@ public BlobStore createInstance(TileLayerDispatcher layers, LockProvider lockPro
...
+        S3BlobStoreInfo blobStore = SerializationUtils.clone(this);
+
+        if (allowEnvParametrization && gwcEnvironment != null) {
+            blobStore.setName(getName());
+            blobStore.setEnabled(isEnabled());
+            blobStore.setDefault(isDefault());
+            blobStore.setAccess(getAccess());
+            blobStore.setPrefix(getPrefix());
+            blobStore.setUseHTTPS(isUseHTTPS());
+            blobStore.setUseGzip(isUseGzip());
```
aaime, line 356: I see these cannot be parameterized... on one side, they seem less likely to be parameterized, on the other end, while you're here, why not do them all?

geowebcache/s3storage/src/main/java/org/geowebcache/s3/S3BlobStoreInfo.java
```diff
@@ -334,7 +337,38 @@ public BlobStore createInstance(TileLayerDispatcher layers, LockProvider lockPro
...
+        return new S3BlobStore(this.clone(gwcEnvironment, true), layers, lockProvider);
+    }
+
+    public S3BlobStoreInfo clone(
+            GeoWebCacheEnvironment gwcEnvironment, Boolean allowEnvParametrization) {
+        S3BlobStoreInfo blobStore = SerializationUtils.clone(this);
+
+        if (allowEnvParametrization && gwcEnvironment != null) {
+            blobStore.setName((String) gwcEnvironment.resolveValue(getName()));
+            blobStore.setEnabled((Boolean) gwcEnvironment.resolveValue(isEnabled()));
```
aaime, line 351: Hum... are you sure the environment will provide back Integer, Access or Integer? I don't see how it can determine the desired type. Most of the env parametrization in GWC needs strings, so the other examples are just fine casting to String (which is, I believe, what you'll actually get back).

Can you please modify the tests in order to test parametrization of all types of properties, so that we can be sure it's actually working?

Also, in order to host parametric values, you'll need to change all the properties in the INfo to be String, I don't see how you can make them parametric otherwise.

geowebcache/s3storage/src/main/java/org/geowebcache/s3/S3BlobStoreInfo.java
```diff
@@ -334,7 +337,38 @@ public BlobStore createInstance(TileLayerDispatcher layers, LockProvider lockPro
...
+            blobStore.setDefault(isDefault());
+            blobStore.setAccess(getAccess());
+            blobStore.setPrefix(getPrefix());
+            blobStore.setUseHTTPS(isUseHTTPS());
+            blobStore.setUseGzip(isUseGzip());
+
+            blobStore.setBucket((String) gwcEnvironment.resolveValue(getBucket()));
+            blobStore.setAwsAccessKey((String) gwcEnvironment.resolveValue(getAwsAccessKey()));
+            blobStore.setAwsSecretKey((String) gwcEnvironment.resolveValue(getAwsSecretKey()));
+            blobStore.setMaxConnections((Integer) gwcEnvironment.resolveValue(getMaxConnections()));
```
aaime, line 361: I thought you reverted back to parameterizing only strings, but this is an Integer instead.

geowebcache/s3storage/src/main/java/org/geowebcache/s3/S3BlobStoreInfo.java
```diff
@@ -334,7 +337,38 @@ public BlobStore createInstance(TileLayerDispatcher layers, LockProvider lockPro
...
+            blobStore.setProxyDomain((String) gwcEnvironment.resolveValue(getProxyDomain()));
+            blobStore.setProxyWorkstation(
+                    (String) gwcEnvironment.resolveValue(getProxyWorkstation()));
+            blobStore.setProxyHost((String) gwcEnvironment.resolveValue(getProxyHost()));
+            blobStore.setProxyPort((Integer) gwcEnvironment.resolveValue(getProxyPort()));
```
aaime, line 366: Here as well.

geowebcache/s3storage/src/main/java/org/geowebcache/s3/S3BlobStoreInfo.java
shadab-geo: ahh, missed it, have updated code

PR #1086, opened 2022-09-15
GWC WMTS data security improvements

davidblasby:
See Geoserver-side PR: https://github.com/geoserver/geoserver/pull/6185
See bug report: https://osgeo-org.atlassian.net/browse/GEOS-10648

When GS global services are off allow GWC to work
When a Layer is non-visible to user, do not show it in the WMTS GetCapabilities
Better handling of workspace-specific requests (instead of global requests).

Changes
1. Smarter normalizeURL (in GeoWebCAcheDispatcher) and getFileName (in ByteStreamController for serving static files) 
2. added JS comment so that Chrome would make it easier to put in breakpoints 
3. Added security check for WMTSGetCapabilites to make sure the user can see the layer before putting it in the document.  This required injecting the SecurityDispatcher.  Updates test cases to inject a null SecurityDispatcher.

### DIFF
--- a/geowebcache/wmts/src/main/java/org/geowebcache/service/wmts/WMTSGetCapabilities.java
+++ b/geowebcache/wmts/src/main/java/org/geowebcache/service/wmts/WMTSGetCapabilities.java
@@ -429,7 +429,7 @@ public class WMTSGetCapabilities {
 
     private void contents(XMLBuilder xml) throws IOException {
         xml.indentElement("Contents");
-        Iterable<TileLayer> iter = tld.getLayerList();
+        Iterable<TileLayer> iter = tld.getLayerListFiltered();
         Set<GridSet> usedGridsets = new HashSet<>();
         for (TileLayer layer : iter) {
             if (!layer.isEnabled() || !layer.isAdvertised()) {
--- a/geowebcache/core/src/main/java/org/geowebcache/GeoWebCacheDispatcher.java
+++ b/geowebcache/core/src/main/java/org/geowebcache/GeoWebCacheDispatcher.java
@@ -226,6 +226,32 @@ public class GeoWebCacheDispatcher extends AbstractController {
         }
     }
 
+    // "/geoserver/topp/gwc/demo/topp:states" --> "/demo/topp:states"
+    //    (workspace "topp" would have been put in LocalWorkspace already)
+    // "/geoserver/gwc/demo/topp:states" --> "/demo/topp:states"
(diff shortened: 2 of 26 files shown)


geowebcache/wmts/src/main/java/org/geowebcache/service/wmts/WMTSGetCapabilities.java
```diff
@@ -73,14 +74,25 @@ public class WMTSGetCapabilities {
 
     private final Collection<WMTSExtension> extensions;
 
+    private SecurityDispatcher securityDispatcher;
```
aaime, line 77: What's a SecurityDispatcher? It's not included in the PR.

Thinking out loud, why don't you work on the TileLayerDispatcher instead, it's sort of a Catalog equivalent, its contents might be filtered on the fly based on a... maybe a pluggable TileLayerDispatcherFilter (similar to the CatalogFilter), that would avoid changing the WMTS code and would also make it support all tiled layer services, rather than having WMTS respond one way, and TMS, WMS-C, VE and GMAPS in another...

davidblasby: I like that idea;

a) will create a TileLayerDispatcherFilter interface
b) provide implementation- as per the code in WMTSGetCapabilities
c) add filtering support in TileLayerDispatcher#getLayers()

davidblasby: thanks!

davidblasby: Just getting ready to commit this.  
I had to add a `getLayersFiltered()` method, since the API contract was to include all layer.  There was code (i.e. for attaching/removing listeners as well as other) that assumed it was getting all the layers.

davidblasby: committed and added test cases

jodygarnett: Looks like it is a much better approach, marking conversation as resolved.

jodygarnett:
Feedback addressed, as discussed in meeting this needs to be merged at the same time as https://github.com/geoserver/geoserver/pull/6185

Issue #1091, opened 2022-09-23
Disk Quota example causes "end tag name </value> must be the same as start tag <units>" error


BrianBaublitz:
The example Disk Quote XML from the [documentation](https://www.geowebcache.org/docs/current/configuration/diskquotas.html) causes an "org.xmlpull.v1.XmlPullParserException: end tag name `</value>` must be the same as start tag `<units>`" exception because of a tag mismatch.

PR #1092, opened 2022-09-23
Fixed end tag on units.

BrianBaublitz:
This patch is in reference to issue [1091](https://github.com/GeoWebCache/geowebcache/issues/1091).


aaime:
Yup, thanks!

PR #1111, opened 2023-01-16
[GWC-1101] Upgrade Spring Core to 5.3.23

turingtestfail:
[GWC-1101] Upgrade Spring Core to 5.3.23

restored tests

cleanup

more test fixes

mimetype is binary check for content type

Update geowebcache/core/src/main/java/org/geowebcache/mime/ApplicationMime.java

Co-authored-by: Andrea Aime <andrea.aime@gmail.com>

constant renaming


smithkm (approved):
Looks good

aaime:
@turingtestfail could you align the version numbers with the GeoServer PR, so that we use the same version in both projects?

turingtestfail:
> @turingtestfail could you align the version numbers with the GeoServer PR, so that we use the same version in both projects?

Done.

PR #1116, opened 2023-02-02
documentation build, site and package

jodygarnett:
See documetation/en/README.md for environment setup and build instructions.


jodygarnett:
@tbarsballe is this what you need to stage documentation into a zip for deploy on osgeo servers?

documentation/en/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+To configure this environment with ``sphinx-build`` (and any other ``requirements.txt``):
+```
+pip install -r requirements.txt
+```
+
+To confirm installation:
+```
+sphinx-build --version
+```
```
tbarsballe, line 39: These are inconsistent with the [GeoServer sphinx install instructions](https://docs.geoserver.org/latest/en/docguide/install.html), which, notably, don't use a venv to install sphinx. 

I think the appropriate change is probably to update the GeoServer instructions, but at the moment, if anyone has set up an environment to build GeoServer docs already, these instructions are going to result in an extra venv.

documentation/en/README.md
```diff
@@ -0,0 +1,39 @@
+# GeoWebCache Documentation
+
+## Building
+
+An ant build script is required (note the python environment described below is required).
```
tbarsballe, line 5: An overview of the produced artifacts (e.g. target/html and *-docs.zip) would be nice here. Looking at the build.xml, they seem pretty consistent with other projects like GeoTools and GeoServer, but it'd still be a good addition, for completeness.

tbarsballe (commented):
Yeah, this looks like what I need. 
I left a few suggestions for improvements to the README, but the actual build.xml looks good to me.

documentation/en/README.md
jodygarnett: Yes I think so; if I understand correctly the requirements.txt is useful with pip (no venv required)

jodygarnett: Fair enough, this technically would allow us to specify the version of sphinx supported/tested by *this* build.

I am going to reword so that the venv is optional.

tbarsballe:
I've created the [docbuild on jenkins](https://build.geoserver.org/view/geowebcache/job/geowebcache-main-live-docs/), and docs are now available at https://geowebcache.osgeo.org/docs/main
There's just a build for main right now (I assume this PR hasn't been backported yet)


We are also still missing a [homepage](https://geowebcache.osgeo.org/). I took a look at https://www.geowebcache.org/ (and downloaded a copy), but it really needs some updating before its useable (not to mention, it could do with being added to version control, similar to [how we've done for GeoServer](https://github.com/geoserver/geoserver/pull/5264))

I've added an issue to capture this remaining work: https://github.com/GeoWebCache/geowebcache/issues/1126

jodygarnett:
Excellent result; there is this in version control https://github.com/GeoWebCache/geowebcache.github.io but I do not know if it is useful.

PR #1118, opened 2023-02-03
[GWC-1117] GWC S3 Blobstore Parameters Get Converted back to plain te…

turingtestfail:
…xt after GeoServer restart

unit test for xstream


geowebcache/s3storage/src/main/java/org/geowebcache/s3/S3BlobStoreConfigProvider.java
```diff
@@ -96,10 +110,6 @@ public XStream getConfiguredXStream(XStream xs) {
         xs.registerLocalConverter(
                 S3BlobStoreInfo.class, "useGzip", EnvironmentNullableBooleanConverter);
         xs.registerLocalConverter(S3BlobStoreInfo.class, "bucket", EnvironmentStringConverter);
```
aaime, line 112: I believe we want to preserve variables for everything. If these converters are expanding on read, they should all be removed, the expansion should be done "on usage" rather than while reading. Same for the ones below.

aaime (changes requested):

geowebcache/s3storage/src/main/java/org/geowebcache/s3/S3BlobStoreConfigProvider.java
turingtestfail: fixed

aaime:
Tested locally with downstream, everything builds.

PR #593, opened 2018-02-16
Configuration API change and REST API Improvements

tbarsballe:
As discussed: 
* https://sourceforge.net/p/geowebcache/mailman/message/36142401/
* https://sourceforge.net/p/geowebcache/mailman/message/36218177/

PR #1124, opened 2023-02-24
GEOS-10006 Seeding GWC doesn't work for layers with a dot in the name

Mike7311:
JIRA: https://osgeo-org.atlassian.net/browse/GEOS-10006

Seeding layers with dot in name from GeoServer UI parsed layer name incorrectly
Modified get mapping to accept layers with dot in name and added additional to process xml
Added test to check if get mapping works correctly with layers which have dot in name

### DIFF
--- a/geowebcache/rest/src/main/java/org/geowebcache/rest/controller/SeedController.java
+++ b/geowebcache/rest/src/main/java/org/geowebcache/rest/controller/SeedController.java
@@ -68,15 +68,23 @@ public class SeedController {
 
     /** GET method for querying running tasks for the provided layer */
     @RequestMapping(
-            value = "/seed/{layer}.json",
+            value = "/seed/{layer:.+}.json",
             method = RequestMethod.GET,
             produces = {MediaType.APPLICATION_JSON_VALUE, MediaType.TEXT_PLAIN_VALUE})
-    public ResponseEntity<?> doGet(HttpServletRequest req, @PathVariable String layer) {
-        return seedService.getRunningLayerTasks(req, layer);
+    public ResponseEntity<?> doGet(@PathVariable String layer) {
+        return seedService.getRunningLayerTasks(layer);
+    }
+
+    @RequestMapping(
+            value = "/seed/{layer:.+}.xml",
+            method = RequestMethod.GET,
+            produces = {MediaType.APPLICATION_XML_VALUE, MediaType.TEXT_PLAIN_VALUE})
+    public ResponseEntity<?> doGetXml(@PathVariable String layer) {
+        return seedService.getRunningLayerTasksXml(layer);
     }
 
     /** GET method for displaying the GeoWebCache UI form. */
-    @RequestMapping(value = "/seed/{layer}", method = RequestMethod.GET)
+    @RequestMapping(value = "/seed/{layer:.+}", method = RequestMethod.GET)
     public ResponseEntity<?> doFormGet(HttpServletRequest request, @PathVariable String layer) {
         return formService.handleGet(request, layer);
     }
--- a/geowebcache/rest/src/main/java/org/geowebcache/rest/service/SeedService.java
+++ b/geowebcache/rest/src/main/java/org/geowebcache/rest/service/SeedService.java
@@ -79,42 +79,57 @@ public class SeedService {
             XStream xs = new GeoWebCacheXStream(new JsonHierarchicalStreamDriver());
             long[][] list = seeder.getStatusList();
             JSONObject obj = new JSONObject(xs.toXML(list));
-            HttpHeaders headers = new HttpHeaders();
-            headers.setContentType(MediaType.APPLICATION_JSON);
-            return new ResponseEntity<>(obj.toString(), headers, HttpStatus.OK);
+            return constructResponse(MediaType.APPLICATION_JSON, obj.toString(), HttpStatus.OK);
         } catch (JSONException jse) {
-            HttpHeaders headers = new HttpHeaders();
-            headers.setContentType(MediaType.TEXT_PLAIN);
-            return new ResponseEntity<Object>("error", headers, HttpStatus.INTERNAL_SERVER_ERROR);
+            return constructResponse(
+                    MediaType.TEXT_PLAIN, "error", HttpStatus.INTERNAL_SERVER_ERROR);
         }
     }
 
     /** GET method for querying running tasks for the provided layer */
-    public ResponseEntity<?> getRunningLayerTasks(HttpServletRequest request, String layer) {
+    public ResponseEntity<?> getRunningLayerTasks(String layer) {
         try {
             XStream xs = new GeoWebCacheXStream(new JsonHierarchicalStreamDriver());
             long[][] list;
-            if (null == layer) {
-                list = seeder.getStatusList();
-            } else {
-                try {
-                    seeder.findTileLayer(layer);
-                } catch (GeoWebCacheException e) {
-                    HttpHeaders headers = new HttpHeaders();
-                    headers.setContentType(MediaType.TEXT_PLAIN);
-                    return new ResponseEntity<>(e.getMessage(), headers, HttpStatus.BAD_REQUEST);
-                }
-                list = seeder.getStatusList(layer);
+            try {
+                list = findTileLayer(layer);
+            } catch (GeoWebCacheException e) {
+                return constructResponse(
+                        MediaType.TEXT_PLAIN, e.getMessage(), HttpStatus.BAD_REQUEST);
             }
             JSONObject obj = new JSONObject(xs.toXML(list));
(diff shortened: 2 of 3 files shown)


geowebcache/rest/src/main/java/org/geowebcache/rest/controller/SeedController.java
```diff
@@ -75,8 +75,13 @@ public ResponseEntity<?> doGet(HttpServletRequest req, @PathVariable String laye
         return seedService.getRunningLayerTasks(req, layer);
     }
 
+    @RequestMapping(value = "/seed/{layer}.xml", method = RequestMethod.GET)
+    public ResponseEntity<?> doGetXml(HttpServletRequest request, @PathVariable String layer) {
+        return formService.handleGet(request, layer);
+    }
```
aaime, line 81: Tested this bit, it does not seem to be necessary:
* formService.handleGet returns HTML outputs anyways
* seedService.getRunningLayerTasks would seem a better fit but it's hard coded to return only JSON (unbelievable, but true...)

However, there is an integration test that would fail without it, ``RestIntegrationTest.testSeedGetLayerXml``... which is not even checking if the output is XML, but would fail if the result is not a 200.
Possible avenues:
* Keep the code as @Judge73 wrote it
* Remove the .xml controller and that test

@tbarsballe git points at you for that test method... althought it might just be a refactor of some pre-existing code. Do you have a preference?

Of course we could also work towards supporting xml output for the seed endpoint, but that seems harder and unrelated to the objective of this PR.

tbarsballe: Looks like the original changes are coming from https://github.com/GeoWebCache/geowebcache/pull/593 , which is where a number of those endpoints were added in the first place. As such, many of the tests (and for that matter, some of the endpoint implementations) are pretty generic.
In this instance, it looks like we just plugged the REST get directly into the corresponding UI method, which appears to manually generate HTML rather than doing something sensible.

For consistency with the rest of the API (and the documentation for that matter), I'd prefer keeping the XML controller around, even if its still actually returning HTML content for now, and opening a new issue to fix the output to produce XML sometime in the future.
If it was to be removed, the swagger documentation should be updated as well.

geowebcache/rest/src/main/java/org/geowebcache/rest/controller/SeedController.java
```diff
@@ -75,8 +75,13 @@ public ResponseEntity<?> doGet(HttpServletRequest req, @PathVariable String laye
         return seedService.getRunningLayerTasks(req, layer);
     }
 
+    @RequestMapping(value = "/seed/{layer}.xml", method = RequestMethod.GET)
```
tbarsballe, line 78: Does this also support layers with dots in their name, or should the path be `"/seed/{layer:.+}.xml"`

geowebcache/rest/src/test/java/org/geowebcache/rest/controller/SeedControllerTest.java
```diff
@@ -89,12 +91,32 @@ public void testSeedLayerNameWithDots() throws Exception {
...
+
+    private void testGet(String layerName) throws Exception {
+        doReturn(ResponseEntity.ok(null))
+                .when(formService)
+                .handleGet(any(HttpServletRequest.class), anyString());
+
+        mockMvc.perform(
+                        get("/rest/seed/{layer}", layerName)
+                                .contentType(MediaType.TEXT_PLAIN)
+                                .accept(MediaType.TEXT_HTML))
```
tbarsballe, line 117: Since XML is now going through a different method, a seperate XML test for dots in name may be warranted, per the question above

geowebcache/rest/src/main/java/org/geowebcache/rest/controller/SeedController.java
tbarsballe: ... and if so, the json endpoint above probably needs the same change too.

geowebcache/rest/src/test/java/org/geowebcache/rest/controller/SeedControllerTest.java
Judge73: Added tests for that

> Since XML is now going through a different method, a seperate XML test for dots in name may be warranted, per the question above

Added tests

geowebcache/rest/src/main/java/org/geowebcache/rest/service/SeedService.java
```diff
@@ -94,27 +91,52 @@ public ResponseEntity<?> getRunningLayerTasks(HttpServletRequest request, String
...
+                list = findTileLayer(layer);
+            } catch (GeoWebCacheException e) {
+                return getBadRequestResponse(
+                        MediaType.TEXT_PLAIN, e.getMessage(), HttpStatus.BAD_REQUEST);
             }
             JSONObject obj = new JSONObject(xs.toXML(list));
-            HttpHeaders headers = new HttpHeaders();
-            headers.setContentType(MediaType.APPLICATION_JSON);
-            return new ResponseEntity<>(obj.toString(), headers, HttpStatus.OK);
+            return getBadRequestResponse(MediaType.APPLICATION_JSON, obj.toString(), HttpStatus.OK);
```
aaime, line 101: I don't know if it was intentional or a mistake, but it seems strange to get a valid result (JSONObject obj) and then pass it through a getBadRequestResponse method. I guess the method is not misnamed, it should be getStringResponse or something similar.

geowebcache/rest/src/main/java/org/geowebcache/rest/service/SeedService.java
```diff
@@ -94,27 +91,52 @@ public ResponseEntity<?> getRunningLayerTasks(HttpServletRequest request, String
...
+            XStream xs = new GeoWebCacheXStream();
+            long[][] list;
+            try {
+                list = findTileLayer(layer);
+            } catch (GeoWebCacheException e) {
+                return getBadRequestResponse(
+                        MediaType.TEXT_PLAIN, e.getMessage(), HttpStatus.BAD_REQUEST);
+            }
+            return getBadRequestResponse(MediaType.APPLICATION_XML, xs.toXML(list), HttpStatus.OK);
+        } catch (JSONException jse) {
```
aaime, line 120: How does one get a JSONException when encoding XML? I believe the code in the JSON method could have thrown it while doing the JSONObject parse and re-encode, but I don't see it here.

geowebcache/rest/src/main/java/org/geowebcache/rest/service/SeedService.java
```diff
@@ -94,27 +91,52 @@ public ResponseEntity<?> getRunningLayerTasks(HttpServletRequest request, String
...
+            LOG.log(Level.SEVERE, jse.getMessage(), jse);
+            return getBadRequestResponse(
+                    MediaType.TEXT_PLAIN, "error", HttpStatus.INTERNAL_SERVER_ERROR);
+        }
+    }
+
+    private ResponseEntity<?> getBadRequestResponse(
+            MediaType textPlain, String message, HttpStatus badRequest) {
```
aaime, line 128:
```suggestion
    private ResponseEntity<?> getResponse(
            MediaType mediaType, String message, HttpStatus status) {
```

Make method and parameter names more generic, at a first sight this method seems to be designed to return an error, but in fact it can return any response whose body is textual.

aaime (changes requested):

geowebcache/rest/src/main/java/org/geowebcache/rest/service/SeedService.java
Judge73: Sorry. Yes, it was a mistake because I was fixing this in parralel with another task and failed to minimize context switch, thus lost some attention. Fixed name.

geowebcache/rest/src/main/java/org/geowebcache/rest/service/SeedService.java
Judge73: Done (same reason)

geowebcache/rest/src/main/java/org/geowebcache/rest/service/SeedService.java
Judge73: Done

aaime:
Tested, I can seed a layer with dots in the name, and get both a xml and a json representation of the ongoing seed.
The JSON representation is what it used to be:

image

The XML one mirrors is:

image

It's not the greatest for sure, but not a regressions either, and changing the JSON one would imply a backwards compatibility so.... good to go IMHO.  Something to keep in mind if one day we'll run a v2 REST API for GWC.

aaime (approved):

jodygarnett:
Thanks @Mike7311 and congrats on your first contribution to GeoWebCache

PR #1147, opened 2023-06-21
Startup DiskQuotaMonitor on context init, rather than after bean properties are set

jodygarnett:
This makes spring initialization much easier and avoids risk of xml configuration not yet being available.

See https://osgeo-org.atlassian.net/browse/GEOS-11032 for background information.


jodygarnett:
The downstream geoserver build is failing on gs-metadata (a common problem) so I believe this PR is ready for review.

aaime:
Integration build restarted, downstream should be sane now, let's see.

Issue #1157, opened 2023-09-22
Truncate on S3BlobStore is not working when diskquota is disabled (no listeners)


JeroenVu:
Hi

We're using S3 storage for our tiles and noticed that old tiles were not being updated even though we did a bounding box truncate. On S3 these tiles effectively didn't get deleted. The truncate job started, ran, and finished without any errors. However, with diskquota enabled, we saw that the files on S3 did get deleted.

I debugged this situation and found that the tile path was appended to the prefix where the prefix didn't have an ending slash. 

So, instead of .../gridset-Lambert72/png/**default/10**/350/354.png
The key became ../gridset-Lambert72/png/**default10**/350/354.png

The DeleteObjectsRequest was totally fine with that and ran without errors and without deleting anything.

I'll try to get this fixed soon.

JeroenVu:
I found a test that fails exactly on this issue, but it has been disabled.

https://github.com/GeoWebCache/geowebcache/blob/ebaddfe67e49d1987b55374aba8e799787ed4308/geowebcache/s3storage/src/test/java/org/geowebcache/s3/AbstractS3BlobStoreIntegrationTest.java#L427

PR #1162, opened 2023-09-25
#1157: Fix for S3BlobStore bulk delete (without listeners)

JeroenVu:
(failing since a133c446, test ignored since 855ed094)

### DIFF
--- a/geowebcache/s3storage/src/main/java/org/geowebcache/s3/S3BlobStore.java
+++ b/geowebcache/s3storage/src/main/java/org/geowebcache/s3/S3BlobStore.java
@@ -311,7 +311,7 @@ public class S3BlobStore implements BlobStore {
     @Override
     public boolean delete(final TileRange tileRange) throws StorageException {
 
-        final String coordsPrefix = keyBuilder.coordinatesPrefix(tileRange, false);
+        final String coordsPrefix = keyBuilder.coordinatesPrefix(tileRange, true);
         if (!s3Ops.prefixExists(coordsPrefix)) {
             return false;
         }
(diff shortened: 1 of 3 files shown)


geowebcache/s3storage/src/main/java/org/geowebcache/s3/S3BlobStore.java
```diff
@@ -311,7 +311,7 @@ public KeyVersion apply(long[] loc) {
     @Override
     public boolean delete(final TileRange tileRange) throws StorageException {
 
-        final String coordsPrefix = keyBuilder.coordinatesPrefix(tileRange, false);
+        final String coordsPrefix = keyBuilder.coordinatesPrefix(tileRange, true);
```
aaime, line 314: The change looks legit, not sure why it was working with the listener thought?
Also, can you add a test to cover this issue?

JeroenVu: Hi Andrea, thanks for your review.

It was working with the listener because the `coordsPrefix` is not used there (see lines 349 and on). 
`if (listeners.isEmpty()) {
<not working, because of coordsPrefix without ending slash>
} else { <working, not using coordsPrefix> }`

https://github.com/geowebcache/geowebcache/blob/a1a70d33facd19445e16abab6f12b35995b62d41/geowebcache/s3storage/src/main/java/org/geowebcache/s3/S3BlobStore.java#L349-L364

The test exists, but has been ignored by https://github.com/GeoWebCache/geowebcache/commit/855ed094b9269a3b919c34aac8e05aff05eec54c.
https://github.com/GeoWebCache/geowebcache/blob/b4dff9c2244fe4569319fe82b453f4f82387fd5f/geowebcache/s3storage/src/test/java/org/geowebcache/s3/AbstractS3BlobStoreIntegrationTest.java#L427

It effectively fails before the change above, and succeeds after the change.

Maybe you want to try to un-ignore this test?

JeroenVu: Done.

PR #1174, opened 2023-10-10
[GWC-1172] Improve handling special characters in the GWC Seed Form

sikeoka:
https://github.com/GeoWebCache/geowebcache/issues/1172
This PR updates the GWC Seed Form to properly escape special characters from GeoServer catalog information.


aaime:
LGTM

PR #1208, opened 2024-01-05
Allow building with QA on Java 17+

groldan:
Running `mvn ... -Dqa` with Java 17/21 breaks errorprone, as in:

```
[INFO] --- compiler:3.8.0:compile (default-compile) @ gwc-core ---
[INFO] Compiling 201 source files to /home/groldan/git/geowebcache/geowebcache/core/target/classes
An exception has occurred in the compiler (17.0.8). Please file a bug against the Java compiler via the Java bug reporting page (https://bugreport.java.com) after checking the Bug Database (https://bugs.java.com) for duplicates. Include your program, the following diagnostic, and the parameters passed to the Java compiler in your report. Thank you.
java.lang.IllegalAccessError: class com.google.errorprone.BaseErrorProneJavaCompiler (in unnamed module @0x3d3e9163) cannot access class com.sun.tools.javac.api.BasicJavacTask (in module jdk.compiler) because module jdk.compiler does not export com.sun.tools.javac.api to unnamed module @0x3d3e9163
        at com.google.errorprone.BaseErrorProneJavaCompiler.addTaskListener(BaseErrorProneJavaCompiler.java:95)
...
````

Fixed with:

* Update errorProne.version: `2.18.0` -> `2.24.1`
* Add `geowebcache/.mvn/jvm.config` with the required `--add-exports` for the `maven-compiler-plugin` as explained at https://errorprone.info/docs/installation#jdk-16.

---

## Fixes to build with Java 21:

* `Thread.getId()` is deprecated. Since it's only used for reporting  errors, use `Thread.getName()` instead
* `new URL(String)` is deprecated. Use `new URI(String).toURL()`  instead, throwing `MalformedURLException` to preserve the old  behavior in methods that throw `IOException`. This is centralized in the new utility class/method `org.geowebcache.util.URLs.of(String url)`
* `ExecutorService` implements `AutoCloseable` in Java 21, add  `@SuppressWarnings("PMD.CloseResource")` in the test classes that use it
* Add `@SuppressFBWarnings` for new rules related to POJO fields unset/null, when they're assigned using reflection by XStream

### DIFF
--- a/geowebcache/core/src/main/java/org/geowebcache/config/XMLConfiguration.java
+++ b/geowebcache/core/src/main/java/org/geowebcache/config/XMLConfiguration.java
@@ -88,6 +88,7 @@ import org.geowebcache.storage.DefaultStorageFinder;
 import org.geowebcache.storage.UnsuitableStorageException;
 import org.geowebcache.util.ApplicationContextProvider;
 import org.geowebcache.util.ExceptionUtils;
+import org.geowebcache.util.URLs;
 import org.springframework.beans.factory.InitializingBean;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Qualifier;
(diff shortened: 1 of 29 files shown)

### CODE REVIEW

File: geowebcache/core/src/main/java/org/geowebcache/config/XMLConfiguration.java [hunk 1]
```diff
@@ -266,13 +268,13 @@ public void setDefaultValues(TileLayer layer) {
             URL proxyUrl = null;
             try {
                 if (getGwcConfig().getProxyUrl() != null) {
-                    proxyUrl = new URL(getGwcConfig().getProxyUrl());
+                    proxyUrl = new URI(getGwcConfig().getProxyUrl()).toURL();
                     log.fine("Using proxy " + proxyUrl.getHost() + ":" + proxyUrl.getPort());
```
aaime, line 272: If I guess correctly, this change is related to the deprecation or "new URL(string)" [in Java 21](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/net/URL.html#%3Cinit%3E(java.lang.String)).

Rather than repeating this pattern over and over, along with the exception rethrowing, could we have a dedicated support method, similar (but not same) to what we have URLs.fileToURL() in GeoTools?

groldan: You got it.

aaime:
Looks good to me, merging.

Issue #1226, opened 2024-03-07
MemoryLockProvider can cause deadlocks if using nested locks


mbosecke:
Because MemoryLockProvider uses striped locking, it's possible for two unrelated keys to collide and block. I understand that this was likely done to reduce memory usage but I'm going to make two arguments towards removing the striping and making it unbounded in size:

1. For this project, I think it's fair to say that performance is one of the top priorities for users, both latency and throughput. Using striped locking will inherently put an upper bound on the throughput ([currently hard-coded to 1024](https://github.com/GeoWebCache/geowebcache/blob/main/geowebcache/core/src/main/java/org/geowebcache/locks/MemoryLockProvider.java#L36)) and will cause random and unpredictable latency when there's contention between unrelated keys. It's also very hard to find the root cause when it happens. I'm working with machines that hundreds of gigabytes of memory and hundreds of CPU cores, but this lock provider has put measurable constraints on performance. I'd say that if a user opts-in to using a "MemoryLockProvider" then it's fair for them to assume that it will come with a small cost of memory, and I think the cost of a single ReentrantLock per-thread is more than acceptable.

2. The provider will cause deadlocks if it's used for nested locks, with no way to defend against it. I noticed this while implementing some changes in Geoserver ([see here](https://github.com/geoserver/geoserver/pull/7457)). Here's an example where none of the locks share keys, and they are acquired/released in a deterministic order,  yet a deadlock will occur if the striping is just unfortunate enough:

```
// Thread 1:
Lock a = getLock("a");
try {
   Lock b = getLock("b"); // BLOCKED BY LOCK C because of striping collision
   try {
        // ... do work
    } finally {
        b.release();
    }
} finally {
   a.release();
}

// Thread 2:
Lock c = getLock("c");
try {
   Lock d = getLock("d"); // BLOCKED BY LOCK A because of striping collision
   try {
        // ... do work
    } finally {
        d.release();
    }
} finally {
   c.release();
}
```

aaime:
Performance is high on the list, for sure, but without sacrificing stability, as these projects are typically running for months at a time. All production setups my company manages always have request limits set up, so that no single request can use more than X memory, and that no more than Y requests can run in parallel (GeoServer service limits, along with control-flow plugins installed), so that one can match the memory given to the JVM.

mbosecke:
[...]

PR #1227, opened 2024-03-07
Changed MemoryLockProvider to be unbounded and support nested locking

mbosecke:
Solves #1226

### DIFF
--- a/geowebcache/core/src/main/java/org/geowebcache/locks/MemoryLockProvider.java
+++ b/geowebcache/core/src/main/java/org/geowebcache/locks/MemoryLockProvider.java
@@ -14,12 +14,12 @@
  */
 package org.geowebcache.locks;
 
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.locks.ReentrantLock;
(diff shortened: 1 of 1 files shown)


geowebcache/core/src/main/java/org/geowebcache/locks/MemoryLockProvider.java
```diff
@@ -14,12 +14,12 @@
...
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.locks.ReentrantLock;
 import java.util.logging.Level;
 import java.util.logging.Logger;
-import org.apache.commons.codec.digest.DigestUtils;
 import org.geotools.util.logging.Logging;
-import org.geowebcache.GeoWebCacheException;
 
 /**
  * An in memory lock provider based on a striped lock
```
aaime, line 25: Please update this javadoc with a summary of how this class works now (some details about lack of bouds, how the map is not actually growing to use infinite memory, and the nested locking).

aaime (commented):
Overall looks good, javadocs needs to be updatd.
I've found some issues in the GeoServer PR, going to wait to be able to run an end-to-end test interactive before I merge this one too.

PR #1232, opened 2024-04-18
Avoid infinite number of calls to listBlobs when doing prefix removals (e..g, gridset or layer removals)

aaime:
I believe this one happened because of a Jackson behavioral change. The code use to check for the "next page" marker in a paged listing to be null, but it's getting an empty string instead. The object representing the response is being deserialized using jackson.

While I was at it, I've fixed also a couple of other issues in tests, which might be also due to behavioral changes, this time in Azure itself (a "a//b" path is not treated as "a/b" but as "a/<emtpyNameFolder/b").


geowebcache/azureblob/src/test/java/org/geowebcache/azure/AbstractAzureBlobStoreIntegrationTest.java
```diff
@@ -154,7 +157,7 @@ public void testPutWithListener() throws MimeException, StorageException {
                         eq(tile.getLayerName()),
                         eq(tile.getGridSetId()),
                         eq(tile.getBlobFormat()),
-                        anyString(),
+                        anyStringOrNull(),
```
aaime, line 160: These are changes in Mockito behavior, previous version was matching null too, the current one does not.

geowebcache/azureblob/src/main/java/org/geowebcache/azure/DeleteManager.java
```diff
@@ -281,12 +291,10 @@ public Long call() throws Exception {
...
                     deleteItems(container, response.body().segment(), filter);
                     String marker = response.body().nextMarker();
-                    if (marker != null) {
-                        response =
-                                container.listBlobsFlatSegment(marker, options, null).blockingGet();
-                    } else {
-                        break;
-                    }
+                    // marker will be empty if there is no next page
+                    if (Strings.isNullOrEmpty(marker)) break;
```
aaime, line 295: This is the actual fix for the high number of listBlob. It would just not get out of this loop.

geowebcache/azureblob/src/main/java/org/geowebcache/azure/DeleteManager.java
```diff
@@ -169,14 +172,21 @@ public void issuePendingBulkDeletes() throws StorageException {
...
                 log.info(
                         String.format(
                                 "Restarting pending bulk delete on '%s/%s':%d",
                                 client.getContainerName(), prefix, timestamp));
-                asyncDelete(prefix, timestamp);
+                if (!asyncDelete(prefix, timestamp)) {
+                    deletesToClear.add(prefix);
+                }
+            }
+            if (!deletesToClear.isEmpty()) {
```
aaime, line 187: Scheduled deletes stored in property files would not get removed if the folder was empty.

dromagnoli (approved):
Looks good to me

PR #1235, opened 2024-04-18
[GWC-1233] Refactor inline JavaScript in the Seed Controller

sikeoka:
This PR moves the inline JavaScript in the Seed Controller into an external file. See https://github.com/GeoWebCache/geowebcache/issues/1233 for more information. This PR is related to Content-Security-Policy work for GeoServer 2.26.0 and should **NOT** be backported.


aaime:
Manually tested, appears to be working correctly. Merging.

Issue #1234, opened 2024-04-18
Refactor inline JavaScript in the Demo Page


sikeoka:
Refactor the Demo Page HTML output to move all inline JavaScript into an external JavaScript file so that the page will work properly with a Content-Security-Policy header that disables unsafe JavaScript. The purpose of this is to reduce the number of exceptions where GeoServer has to set a Content-Security-Policy that enables unsafe JavaScript.

See https://osgeo-org.atlassian.net/browse/GEOS-11346 for more information.

PR #1236, opened 2024-04-18
[GWC-1234] Refactor inline JavaScript in the Demo Page

sikeoka:
This PR moves the inline JavaScript in the Seed Controller into an external file. See https://github.com/GeoWebCache/geowebcache/issues/1234 for more information. This PR is related to Content-Security-Policy work for GeoServer 2.26.0 and should **NOT** be backported.

All Java inputs that were previously written directly into the script tag are now written to hidden input fields that the external JavaScript file will read. The previous Demo Page Java code had to mix HTML and JavaScript escaping depending on where inputs were being written to but the new code only needs to use HTML escaping.

The GetFeatureInfo output was previously handled with a JavaScript XMLHttpRequest call and writing the response into the DOM but is now handled by an iframe like the WMS GetMap OpenLayers3 output format. The reason for this is that the iframe uses its own CSP which allows the GetFeatureInfo HTML to use a more lenient CSP than the Demo page. For example, the GetFeatureInfo CSP can enable inline scripts even if the Demo Page CSP disables inline scripts.

Example of the hidden inputs:
```<input type="hidden" id="hidden_dpi" value="90.71428571428572" />
<input type="hidden" id="hidden_gridsetName" value="EPSG:4326" />
<input type="hidden" id="hidden_gridNames" value="[&quot;EPSG:4326:0&quot;, &quot;EPSG:4326:1&quot;, ..." />
<input type="hidden" id="hidden_format" value="image/png" />
<input type="hidden" id="hidden_layerName" value="ne:boundary_lines" />
<input type="hidden" id="hidden_SRS" value="EPSG:4326" />
<input type="hidden" id="hidden_unit" value="degrees" />
<input type="hidden" id="hidden_resolutions" value="[0.703125, 0.3515625, 0.17578125, 0.087890625, ...]" />
<input type="hidden" id="hidden_tileWidth" value="256" />
<input type="hidden" id="hidden_tileHeight" value="256" />
<input type="hidden" id="hidden_minX" value="-180.0" />
<input type="hidden" id="hidden_maxY" value="90.0" />
<input type="hidden" id="hidden_isVector" value="false" />
<input type="hidden" id="hidden_maxX" value="180.0" />
<input type="hidden" id="hidden_minY" value="-90.0" />
<input type="hidden" id="hidden_fullGrid" value="false" />
<input type="hidden" id="hidden_origin" value="[-180.0, 90.0]" />
<input type="hidden" id="hidden_bbox" value="[-180.0,-90.0,180.0,90.0]" />
<input type="hidden" id="hidden_zoomBounds" value="[-140.99778,-54.89681,141.0338517600138,70.16419]" />```

### DIFF
--- a/geowebcache/core/src/main/java/org/geowebcache/demo/Demo.java
+++ b/geowebcache/core/src/main/java/org/geowebcache/demo/Demo.java
@@ -15,7 +15,6 @@
 package org.geowebcache.demo;
 
 import static org.apache.commons.text.StringEscapeUtils.escapeHtml4;
-import static org.owasp.encoder.Encode.forJavaScript;
 
 import java.io.IOException;
 import java.util.ArrayList;
@@ -50,7 +49,6 @@ import org.geowebcache.mime.ImageMime;
 import org.geowebcache.mime.MimeType;
 import org.geowebcache.mime.XMLMime;
 import org.geowebcache.util.ServletUtils;
-import org.owasp.encoder.Encode;
 import org.springframework.util.Assert;
 
 public class Demo {
@@ -291,7 +289,7 @@ public class Demo {
             formatMime = layer.getDefaultMimeType();
         }
 
-        StringBuffer buf = new StringBuffer();
+        StringBuilder buf = new StringBuilder();
 
         String openLayersPath = "../rest/web/openlayers3/";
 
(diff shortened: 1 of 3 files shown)


aaime:
Tested interactively, appears to be working fine. Merging.

PR #1279, opened 2024-06-09
Updating PMD to 7.2.0

aaime:
Mostly cleaning up unecessary type specs, empty statements and unnecessary casts.


jodygarnett (approved):
This looks good, I like how the code is getting more terse over time.

PR #1284, opened 2024-06-28
Preemptively send authentication credentials in WMSHttpHelper

patlkli:
While trying to cache a WMS layer from another GeoServer instance that requires authentication, I noticed that GeoWebCache didn't manage to perform any successful requests to that upstream GeoServer.

While digging into that issue, I noticed that httpcomponents will only send credentials when challenged by the server.
Since that GeoServer had publicly available layers, it would just respond with 404 instead of 401, so `WMSHttpHelper` wouldn't try it again with credentials.

Before moving from commons-httpclient to httpcomponents in 8beede1, `WMSHttpHelper` was sending authentication credentials preemptively. This behaviour changed after the move.

In fact, httpcomponents doesn't seem to have any mechanism to force preemptive authentication except for explicitly setting the Authorization header. So that's exactly what this PR does.

Just as a side note, `BasicScheme.authenticate` doesn't actually ever throw `AuthenticationException`, that's why it's re-thrown as an `AssertionError` here.

### DIFF
--- a/geowebcache/core/src/main/java/org/geowebcache/layer/wms/WMSHttpHelper.java
+++ b/geowebcache/core/src/main/java/org/geowebcache/layer/wms/WMSHttpHelper.java
@@ -30,11 +30,14 @@ import org.apache.commons.io.IOUtils;
 import org.apache.http.HttpEntity;
 import org.apache.http.HttpResponse;
 import org.apache.http.NameValuePair;
+import org.apache.http.auth.AuthenticationException;
+import org.apache.http.auth.UsernamePasswordCredentials;
 import org.apache.http.client.HttpClient;
 import org.apache.http.client.methods.HttpGet;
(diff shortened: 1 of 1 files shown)


aaime:
@ianturton you worked on the HTTP components upgrade. Any observation about the preemptive authentication change?

jodygarnett:
I am not sure what to with this PR, do we have a test or a downstream integration test or any way to ensure this does something?

ianturton:
I can't really comment on the change - it looks fine but I can't think of how to test it for real unless we have an authenticated server to test against.

aaime:
A test can be written using wiremock, stubbing expected requests and responses. Example here:

https://github.com/geotools/geotools/blob/d3ab583cf5f4cd239cf9fba63834a319905ddfc7/modules/library/http/src/test/java/org/geotools/http/SimpleHttpClientTest.java#L52

PR #1296, opened 2024-07-28
Show vector tiles attributes on hover

aaime:
Vector tiles preview only provides a wireframe, I believe it's important to also show attributes are there, on hover (clicking stops the hover and allows to inspect the attribute list, in case there are too many):

image


groldan (approved):
LGTM!

aaime:
Thanks, merging!

PR #1298, opened 2024-08-07
Fix Azure blobstore error instantiating AzureClient

groldan:
commit 72ce566b "Avoid infinite number of calls to listBlobs when doing prefix removals (e..g, gridset or layer removals)" changed the gwc `AzureClient` check for container existence from an attemp to create it to an attempt to get its properties before creating it.

When the container doesn't exist, the blocking call throws `com.microsoft.azure.storage.blob.StorageException`, which contains the status code.

Wrap that call in a try-catch block and get the status code from the exception.

### DIFF
--- a/geowebcache/azureblob/src/main/java/org/geowebcache/azure/AzureClient.java
+++ b/geowebcache/azureblob/src/main/java/org/geowebcache/azure/AzureClient.java
@@ -93,8 +93,13 @@ class AzureClient implements Closeable {
             this.container = serviceURL.createContainerURL(containerName);
             // no way to see if the containerURL already exists, try to create and see if
             // we get a 409 CONFLICT
+            int status;
             try {
-                int status = this.container.getProperties().blockingGet().statusCode();
+                status = this.container.getProperties().blockingGet().statusCode();
+            } catch (com.microsoft.azure.storage.blob.StorageException se) {
+                status = se.statusCode();
+            }
+            try {
                 if (status == HttpStatus.NOT_FOUND.value()) {
                     status = this.container.create(null, null, null).blockingGet().statusCode();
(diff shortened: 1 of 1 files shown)


geowebcache/azureblob/src/main/java/org/geowebcache/azure/AzureClient.java
```diff
@@ -93,8 +93,13 @@ public AzureClient(AzureBlobStoreData configuration) throws StorageException {
             this.container = serviceURL.createContainerURL(containerName);
             // no way to see if the containerURL already exists, try to create and see if
             // we get a 409 CONFLICT
+            int status;
```
aaime, line 96: Tested interactively, I can't get the code to throw an exception? Different azure container setup maybe? Or a difference in dependencies?

groldan: My first thought was it would be a difference in dependencies, tried everything I could think of, but the only thing that worked was reverting gwc 1.26-SNAPSHOT to 1.25-SNAPSHOT.
So I debugged both and saw the difference, the patch is not in the 1.25.x branch.
BTW the exception happens against the Azurite test container when running integration tests in gscloud

aaime: Ok so you confirm the issue is not happening against an actual Azure storage, but the Azurite test container. Would be nice if the title was clear about it.

groldan: Hi, I actually confirm the issue happens both with Azure and Azurite. The only difference is you may be always running the tests against an Azure container that already exists, whereas the Azurite test has to create it.
If you try a container name that does not exist in Azure you should get the same error.

PR #1300, opened 2024-08-08
Upgrade git-commit-id-plugin to 4.9.0 and set useNativeGit to not fail on git worktrees

groldan:
`jgit` does not support git worktrees. Setting the config option `<useNativeGit>true</useNativeGit>` makes it not fail.


groldan:
In GeoServer one has to run maven with `-Dgit.commit.useNative=true`. It'd be nice if it was the default. I'm not sure how many devs use worktrees, but I use them all the time. In any case, whomever is trying to run a build, most probably have native git installed, though I'm not sure if the plugin would fall back to jgit when it's not?

PR #1303, opened 2024-08-11
Run Azure blob store integration tests with Azure and Azurite

groldan:
Azurite is run as a testcontainer, and the github CI build runs all integration tests for Java 11, 17, and 21.

In order to save on Azure resources, online integration tests are run only if the Azurite based tests succeeded, using github secrets to set up the account, key, and container; and for Java 11 only.

Both Azurite and Azure tests are run by the failsafe-maven-plugin, during the integration-test phase, so you need to run them as `mvn verify`, not as `mvn test`.

### DIFF
--- a/.github/workflows/azure-integration.yml
+++ b/.github/workflows/azure-integration.yml
@@ -1,0 +1,88 @@
+name: Azure BlobStore Integration
+
+on:
+  push:
+    branches:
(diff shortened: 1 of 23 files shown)


groldan:
Closing as superseded by #1304

PR #1304, opened 2024-08-11
Run Azure blob store integration tests with Azure and Azurite

groldan:
Azurite is run as a testcontainer, and the github CI build runs all integration tests for Java 11, 17, and 21.

In order to save on Azure resources, online integration tests are run only if the Azurite based tests succeeded, using github secrets to set up the account, key, and container; and for Java 11 only.

Both Azurite and Azure tests are run by the failsafe-maven-plugin, during the integration-test phase, so you need to run them as mvn verify, not as mvn test.


geowebcache/core/src/test/java/org/geowebcache/storage/AbstractBlobStoreTest.java
```diff
@@ -1129,7 +1133,15 @@ protected void assertNoTile(
...
-        assertThat(store.get(to), describedAs("don't get a tile", is(false)));
+        assertNoTile(to);
+    }
+
+    private void assertNoTile(TileObject to) {
+        await().atMost(5, SECONDS) // give stores with async deletes a chance to complete
+                .untilAsserted(
+                        () ->
+                                assertThat(
+                                        store.get(to), describedAs("don't get a tile", is(false))));
```
groldan, line 1144: Note I had to use `Awaitlity` for the tests in `AbstractBlobStoreTest` that do deletes because the online tests may be too slow to complete async deletes on Azure. Couldn't make them fail on my dev machine, but they failed randomly on github actions

groldan:
[New jobs](https://github.com/GeoWebCache/geowebcache/actions/runs/10336846985):

image

please note the Azure online tests are being run against my free-tier account while I wait for camptocamp to provision an account we can use for testing.

aaime:
It seems all Azure tests are run twice, push vs pull_request?

image

image

.github/workflows/azure-integration.yml
```diff
@@ -0,0 +1,88 @@
...
+      - "geowebcache/azureblob/**"
+  pull_request:
+    branches:
+      - "main"
+    paths:
+      - ".github/workflows/azure-integration.yml"
+      - "pom.xml"
+      - "geowebcache/pom.xml"
+      - "geowebcache/core/**"
+      - "geowebcache/azureblob/**"
```
aaime, line 21: Github actions in GT/GWC/GS normally only have:

```
on: [pull_request]
```

Builds are not run on pushes (too late), and are run on all pull request regardless of the branch.

.github/workflows/linux.yml
```diff
@@ -27,7 +27,8 @@ jobs:
         restore-keys: |
           ${{ runner.os }}-maven-
     - name: Build with Maven
-      run: mvn -B clean install -Dspotless.apply.skip=true -Dall -T2 --file geowebcache/pom.xml
+      # -DskipITs skips failsafe integration tests but runs unit tests with surefire. Integration tests have their own jobs
+      run: mvn -B clean install -Dspotless.apply.skip=true -Dall -T2 --file geowebcache/pom.xml -DskipITs
```
aaime, line 31: The practice in GT/GS/GWC is that online tests are run only when explicitly required. Please don't introduce an opposite practice with a different flag name to enable it (option B, discuss this new approach on list and then apply uniformly in all 3 projects).

geowebcache/azureblob/src/test/java/org/geowebcache/azure/tests/container/AzuriteAzureBlobStoreConformanceIT.java
```diff
@@ -0,0 +1,51 @@
...
+ * Runs {@link AzureBlobStoreConformanceTest} tests against a local ephemeral Docker container using
+ * {@link AzuriteContainer}.
+ *
+ * <p>If there's no Docker environment, the test is {@link AzuriteContainer#disabledWithoutDocker()
+ * ignored}
+ */
+public class AzuriteAzureBlobStoreConformanceIT extends AzureBlobStoreConformanceTest {
+
+    /**
+     * Use "legacy" container to work with {@literal
```
aaime, line 34: Ah ah, that's how you made it work!

geowebcache/core/src/test/java/org/geowebcache/storage/AbstractBlobStoreTest.java
```diff
@@ -959,7 +962,9 @@ public void testDeleteByParametersIdDoesNotDeleteOthers() throws Exception {
         store.put(toCache1);
         store.put(toCache2);
         store.deleteByParametersId("testLayer", paramID1);
-        assertThat(store.get(fromCache2_3), is(true));
+
+        await().atMost(5, SECONDS) // give stores with async deletes a chance to complete
```
aaime, line 966: Github actions are often very unpredictable, time-wise. I would give it a much larger max wait... if I remember correctly, this is the sort of wait that ends the moment the condition is met, and the condition is actually tested much more often.

geowebcache/core/src/test/java/org/geowebcache/storage/AbstractBlobStoreTest.java
aaime: No problem.

geowebcache/pom.xml
```diff
@@ -521,6 +534,28 @@
...
+          </executions>
+          <inherited>true</inherited>
+          <configuration>
+            <systemPropertyVariables>
+              <java.util.logging.config.file>src/test/resources/logging.properties</java.util.logging.config.file>
+              <!-- prevent ForkedBooter focus steal on macOS -->
+              <java.awt.headless>true</java.awt.headless>
+            </systemPropertyVariables>
+            <argLine>${maven.test.jvmargs} -XX:+IgnoreUnrecognizedVMOptions --illegal-access=warn --add-exports=java.desktop/sun.awt.image=ALL-UNNAMED --add-opens=java.base/java.lang=ALL-UNNAMED --add-opens=java.base/java.util=ALL-UNNAMED --add-opens=java.base/java.lang.reflect=ALL-UNNAMED --add-opens=java.base/java.text=ALL-UNNAMED --add-opens=java.desktop/java.awt.font=ALL-UNNAMED</argLine>
+            <skipTests>${skipITs}</skipTests>
```
aaime, line 556: skipITs seems to be [the default way to have failsafe skip running the integration tests](https://maven.apache.org/surefire/maven-failsafe-plugin/examples/skipping-tests.html).

The skipTests parameter is [apparently deprecated](https://maven.apache.org/surefire/maven-failsafe-plugin/integration-test-mojo.html).

geowebcache/azureblob/src/test/java/org/geowebcache/azure/tests/container/AzuriteAzureBlobStoreConformanceIT.java
groldan: yes, and a pinch of blood and sewat

geowebcache/pom.xml
groldan: Meanwhile, `skipTests` is the standard way to skip running the tests but compiling them. There's a reason to set `<skipTests>${skipITs}</skipTests>` in the failsafe configuration. Adding the following comment:
```xml
<!-- allow skipping unit tests but run integration tests with -DskipTests -DskipITs=false, otherwise skipTests will also disable the failsafe run -->
```

.github/workflows/azure-integration.yml
groldan: nice, `branches:*` was a leftover from working on my fork to be able of run them before the pr.
Changed it to run an all pr's regardless of the branch, and on all tags.

geowebcache/core/src/test/java/org/geowebcache/storage/AbstractBlobStoreTest.java
groldan: changed it to 30s

.github/workflows/linux.yml
groldan: but for the traditional `*OnlineTest.java` tests, "explicitly required" means the required `.properties` file is in place.
In gt/gs/gwc, so far nobody uses the failsafe plugin to run integration tests (except gs' `community/rest-openapi/pom.xml` and gt's `modules/unsupported/elasticsearch`.

My point of view is if you wrote `*IT.java` integration tests to be run by failsafe, it is because you expect them to run after the unit tests succeeded and after the maven package phase. And hopefully as often as the unit tests (e.g. during development).

So introducing `-DskipITs` here makes sense cause we want to preserve the traditional behavior [*], and setup specific github actions to run ITs for specific modules (e.g. I'd like to do the same for the S3 blobstore), but still the default action would be to run unit then integration tests if you do `mvn clean install` or `mvn verify` during development.

> \* i.e., run both tests and online tests with surefire

aaime: I see that you're going towards the maven default behavior, which is not bad per-se.
But it's also something that should be discussed on the list (geotools-devel or geoserver-devel). Having slow-to-execute tests run by default is not our default, the community should decide if it's desirable to start with, and how to deal with the (temporary?) inconsistency between existing online tests and the newly introduced integration tests.

groldan: I've no trouble disabling failsafe by default. What you propose would be a separate discussion than the one for this pull request, and I'm afraid I don't have the bandwidth to engage right now. I'd rather spend my time on the actual migration to the new libraries, since it'll also imply doing to at the imageio-ext side of the fence.

But not as it stands, in order to run the traditional "online" tests for Azure, it won't be enough to have the .properties file in place, you'll also need to specify `-DskipITs=false`, since I've refactored them to run during the integration-tests phase. Otherwise you'd be running the online tests at the same time than the unit tests, and hopefully one should avoid that.

Is that a good compromise?

aaime: @groldan in our geo-trio projects we always highlighted the importance of communicating before coding, I see you assuming that what you're doing is "obviously better and thus requires no discussion" too often. 
Since you don't want to have preparatory discussions on the list, we're going to have at least one here, and it can't be just me and you, it needs a few more interested parties.

Since it's PSC meeting day, I'll add the usage of the failsafe plugin as a topic, again for the sake of having a larger audience.

@jodygarnett @smithkm can you please share your opinion here? Personally I'm happy with whatever consistent direction, as long as there is broad agreement on the way to go.

aaime: I've run the build, the ITs add a couple of minutes to the Azure module build.

In terms of controlling the test execution, an existing alternative example s the ElasticSearch tests in GeoTools, which uses failsafe and the IT naming convention, does not need a property file, but has the failsafe execution enabled [only when the online profile is enabled](https://github.com/geotools/geotools/blob/main/modules/unsupported/elasticsearch/pom.xml#L197).

groldan: >@groldan in our geo-trio projects we always highlighted the importance of communicating before coding, I see you assuming that what you're doing is "obviously better and thus requires no discussion" too often.

In this particular case, running the integration tests after the unit tests is not only "obviously better" than not, but also standard practice in the industry.

As for your general statement about my behavior, I strongly disagree and I'm sorry you feel that way.

> Since you don't want to have preparatory discussions on the list, we're going to have at least one here, and it can't be just me and you, it needs a few more interested parties.

A pull request by definition is a proposal for a change, and an opportunity to discuss with the bulk of the work done in advance, and hence the confidence that the proposed approach works, as opposed to discussing in the air.
The pull request contents are way more important than the discussion of whether we run the integration tests by default or not (given I've solved a problem that lasted for years) , for which I agreed to revert to usual behavior, but asked if you're ok with running the online tests when explicitly asked for. The mechanics of it could be through `-DskipITs` or `-Ponline`, but I didn't think the practice on a community module would be any indication of a standard practice.

> I've run the build, the ITs add a couple of minutes to the Azure module build.
I don't get if that's a complain or just a statement

groldan: `-Ponline` it is, as discussed in the PSC meeting. I'll do that.

groldan:
Made it so the ITs only run with `-Ponline`.

aaime (approved):
Had another go at testing the build, looking good, thanks!

PR #1305, opened 2024-08-13
Upgrade hamcrest from 1.3 to 3.0

sikeoka:
This is a test dependency upgrade.

This PR adds the hamcrest 3.0 dependency to the root pom.xml and removes all hamcrest-core, hamcrest-library and hamcrest-integration dependencies. The version is set to 3.0 for hamcrest-core and hamcrest-library in case there is another transitive dependency on it (these are empty jar files). hamcrest-integration has no version beyond 1.3.

Some code changes were required for deprecated or removed API.

### DIFF
--- a/geowebcache/core/src/test/java/org/geowebcache/seed/TruncateBboxRequestTest.java
+++ b/geowebcache/core/src/test/java/org/geowebcache/seed/TruncateBboxRequestTest.java
@@ -28,12 +28,14 @@ import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
 import org.easymock.EasyMock;
+import org.easymock.IArgumentMatcher;
 import org.geowebcache.grid.BoundingBox;
 import org.geowebcache.grid.GridSubset;
 import org.geowebcache.layer.TileLayer;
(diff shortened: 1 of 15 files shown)


geowebcache/core/src/test/java/org/geowebcache/seed/TruncateBboxRequestTest.java
```diff
@@ -58,7 +60,20 @@ protected SeedRequest seedRequest(
...
+                    public boolean matches(Object argument) {
+                        return matcher.matches(argument);
+                    }
+
+                    @Override
+                    public void appendTo(StringBuffer buffer) {
+                        matcher.describeTo(new StringDescription(buffer));
+                    }
+                });
         return null;
```
aaime, line 77: This bit has me confused.. if we return null, is the object being built in the previous step even used?

sikeoka: The old and new code both return null here. I think that easymock is adding the matcher to a static field so the return value isn't very important. I personally find mockito to be clearer and easier to use even though geoserver uses easymock more.

PR #1312, opened 2024-08-20
Upgrade Spring version from 5.3.34 to 5.3.39

mprins:
see also https://github.com/geoserver/geoserver/pull/7845 and https://github.com/geoserver/geoserver/pull/7846 and associated Jira ticket https://osgeo-org.atlassian.net/browse/GEOS-11506


aaime:
Thanks!

PR #1314, opened 2024-08-26
Replace long[] arrays by a TileIndex abstraction

groldan:
Dealing with `long[]` everywhere and knowing it means `{x,y,z}` is tiring.

Introducing a `TileIndex` abstraction improves the semantics, and at the same time allows to save memory on TileObject by storing the coordinates as internal long,long,int fields and saving the object reference to a `long` array.

---

> This is a **proposal**. Can we please stop using `long[]` everywhere?
> If accepted, there would be more places to get rid of `long` arrays
> and introduce higher level abstractions (e.g. `TileRangeInterator`)


groldan:
closing as stale

PR #1322, opened 2024-09-23
Set disk quota store isolation to serializable

aaime:
The default isolation level (read committed) allows for phantom reads and non repeatable reads, which is probably the  reason why the disk quota totals sometimes diverge from reality. Since the updates are asynch, it should not cause major bottlenecks (hopefully).


smithkm (approved):

jodygarnett:
did you wish this to be backported?

PR #1330, opened 2024-10-29
[GWC-1329] Upgrade log4j to 2.24.1 and slf4j to 2.0.16

sikeoka:
This PR upgrades the log4j and slf4j dependencies to their current versions and is related to GeoServer's Wicket 9 upgrade.

See https://osgeo-org.atlassian.net/browse/GEOS-11590 for more information.


aaime:
@groldan is it because of the missing paid-for Azure account? Should we disable the online test and just rely on the Azurite one?

Issue #1325, opened 2024-10-18
geowebcache link in geoserver links to geowebcache.org instead of geowebcache.osgeo.org


m-lmi:
I wanted to report that geowebcache link on http://localhost:8080/geoserver/gwc/home 
![image](https://github.com/user-attachments/assets/fc4f31e2-6d7e-47fa-a901-be1655f68f50)

Is leading to http://geowebcache.org/ which seems not the right website. 
![image](https://github.com/user-attachments/assets/09cced0b-d96b-4fa1-9004-5655a12f98f3)

This should be the correct website: https://geowebcache.osgeo.org/

Also if I check in Google I get this as main page for geowebcache, which seems incorrect. 
![image](https://github.com/user-attachments/assets/32cb504f-ed4e-4090-bd1c-25af08e725b7)

petersmythe:
Thank you for reporting this @m-lmi - much appreciated!

PR #1341, opened 2024-12-29
Update old geowebcache.org links to geowebcache.osgeo.org

petersmythe:
Except for schema URLs e.g. http://geowebcache.org/schema/1.9.0/geowebcache.xsd

Fixes: https://github.com/GeoWebCache/geowebcache/issues/1325

Also reorder GetCapabilities links in http://localhost:8080/geoserver/gwc/home, so that **latter** refers to the correct (WMS) GetCapabilities (and not TMS):

image

### DIFF
--- a/documentation/en/user/source/production/index.rst
+++ b/documentation/en/user/source/production/index.rst
@@ -109,7 +109,7 @@ These environment variables can be established by any of the following ways, in 
 Resource Allocation
 -------------------
 
-Also see http://geowebcache.org/trac/wiki/resources for tools that can be used to estimate how much storage you need and how long seeding will take
+Also see <del>http://geowebcache.org/trac/wiki/resources</del> for tools that can be used to estimate how much storage you need and how long seeding will take
 
 
 Clustering
--- a/documentation/en/user/source/production/index.rst
+++ b/documentation/en/user/source/production/index.rst
@@ -109,7 +109,7 @@ These environment variables can be established by any of the following ways, in 
 Resource Allocation
 -------------------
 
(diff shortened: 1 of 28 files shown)


documentation/en/user/source/production/index.rst
```diff
@@ -109,7 +109,7 @@ These environment variables can be established by any of the following ways, in
 Resource Allocation
 -------------------
 
-Also see http://geowebcache.org/trac/wiki/resources for tools that can be used to estimate how much storage you need and how long seeding will take
+Also see <del>http://geowebcache.org/trac/wiki/resources</del> for tools that can be used to estimate how much storage you need and how long seeding will take
```
petersmythe, line 112: I can't find an update for `/trac/wiki/resources`, so have chosen to strike this out until it can be found.

jodygarnett: I do not think we are going to find that one! But the internet remembers:

http://web.archive.org/web/20100326130953/http://geowebcache.org/trac/wiki/resources

The spreadsheet is lost to time but I added the table to our wiki here https://github.com/GeoWebCache/geowebcache/wiki/Estimating-the-number-of-tiles-and-size-on-disk

jodygarnett, line 112:
```suggestion
Also see https://github.com/GeoWebCache/geowebcache/wiki/Estimating-the-number-of-tiles-and-size-on-disk for table that can be used to estimate how much storage you need and how long seeding will take
```

jodygarnett (changes requested):
Small change to address trac wiki link

documentation/en/user/source/production/index.rst
petersmythe: Wow, thank you!

Issue #1344, opened 2025-01-14
Hide version info on GWC home page


sikeoka:
Pull Request: https://github.com/GeoWebCache/geowebcache/pull/1345

PR #1345, opened 2025-01-14
[GWC-1344] Hide version info on GWC home page

sikeoka:
This PR hides version information and some other internal server information from the GeoWebCache home page unless the user is logged in as an administrator.

Issue: https://github.com/GeoWebCache/geowebcache/issues/1344

### DIFF
--- a/geowebcache/core/src/main/java/org/geowebcache/filter/security/SecurityDispatcher.java
+++ b/geowebcache/core/src/main/java/org/geowebcache/filter/security/SecurityDispatcher.java
@@ -73,6 +73,14 @@ public class SecurityDispatcher implements ApplicationContextAware {
         }
     }
 
+    /**
+     * Checks if the current user is authenticated and is the administrator. Will return true if there are no active
+     * security filters.
+     */
+    public boolean isAdmin() {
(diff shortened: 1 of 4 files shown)


geowebcache/core/src/main/java/org/geowebcache/filter/security/SecurityDispatcher.java
```diff
@@ -73,6 +73,14 @@ public void checkSecurity(TileLayer layer, @Nullable BoundingBox extent, @Nullab
         }
     }
 
+    /**
```
aaime, line 76: As far as I know, if GWC is deployed stand-alone, there are indeed no SecurityFilter, the only implementation I'm aware of is in GeoServer (which you have updated accordingly [in the GS counterpart](https://github.com/geoserver/geoserver/pull/8189)).

I'm asking because I don't want to give the wrong impression in the GWC release notes... the commit message should probably be something along the lines of "Support hiding GWC version when integrated in GeoServer".
(we use the commit messages to build the GWC release notes)

jodygarnett:
@aaime I have addressed your commit message feedback by "embedded GWC" to imply in GeoServer only.

PR #1396, opened 2025-04-21
Mime type: advertise preferred content disposition

aaime:
Simple change to make OGC API Tiles user experience better (decide whether to download a tile, or to display it).


dromagnoli (approved):
Looks good to me.

PR #1459, opened 2025-12-09
FullWMS: support fusing transparent tiles and generating an opaque output

aaime:
In addition I had to comment out yet another Validator usage in tests... the xlink.xsd is not getting downloaded by XMLUnit, even if it's available from the browser. We'll need to rethink that, but for the moment it's mostly a distraction, that class has plenty validators already commented out.

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -45,7 +45,7 @@ jobs:
         cd ~
         mvn -B -f geoserver/src/pom.xml install -nsu -Prelease -Dspotless.apply.skip=true -DskipTests
         mvn -B -f geoserver/src/community/pom.xml install -nsu -DcommunityRelease -Dspotless.apply.skip=true -DskipTests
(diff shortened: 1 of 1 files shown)


geowebcache/wmts/src/test/java/org/geowebcache/service/wmts/WMTSServiceTest.java
```diff
@@ -840,9 +839,11 @@ public void testGetCapEmptyStyleFilter() throws Exception {
 
         String result = resp.getContentAsString();
 
-        Validator validator = new Validator(result);
-        validator.useXMLSchema(true);
-        validator.assertIsValid();
+        System.setProperty("http.agent", "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36");
```
dromagnoli: is it needed?
what about the validators commented out section ?
(wondering if that was part of a debug session)

aaime: Good catch, I was trying to have the validator use a different user agent, as xlink.xsd was available from a browser.

PR #1518, opened 2026-04-28
GWC improvements:better resolution choice and Support HintsLevel

dromagnoli:
Update resolution selection, now based on nearest match and support configuring HintsLevel on Layer definition.

### DIFF
--- a/geowebcache/core/src/main/java/org/geowebcache/layer/wms/WMSLayer.java
+++ b/geowebcache/core/src/main/java/org/geowebcache/layer/wms/WMSLayer.java
@@ -28,6 +28,7 @@ import org.apache.hc.core5.http.ClassicHttpResponse;
 import org.apache.hc.core5.http.HttpEntity;
 import org.geotools.util.logging.Logging;
 import org.geowebcache.GeoWebCacheException;
+import org.geowebcache.config.HintsLevel;
 import org.geowebcache.config.XMLGridSubset;
 import org.geowebcache.config.legends.LegendsRawInfo;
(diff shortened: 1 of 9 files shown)


geowebcache/core/src/main/java/org/geowebcache/layer/wms/WMSLayer.java
```diff
@@ -679,6 +682,15 @@ public String getProxyUrl() {
         return proxyUrl;
     }
 
+    @Override
+    public HintsLevel getHintsLevel() {
+        return hintsLevel;
```
aaime: Should check if hintsLevel is null and do something sensible, like maybe returning the default hints, to be consistent with TileLayer? Would also spare a few null checks in the codebase.

geowebcache/wms/src/main/java/org/geowebcache/service/wms/WMSTileFuser.java
```diff
@@ -294,6 +200,13 @@ protected WMSTileFuser(TileLayerDispatcher tld, StorageBroker sb, HttpServletReq
         reqHeight = Integer.valueOf(values.get("height"));
 
         fullParameters = layer.getModifiableParameters(servReq.getParameterMap(), servReq.getCharacterEncoding());
+        if (layer instanceof WMSLayer) {
```
aaime: Since the hints levels have been added to the TileLayer, there is no need for this check I believe.

aaime (changes requested):

dromagnoli:
All feedbacks have been addressed.

PR #1523, opened 2026-05-05
Do not close JNDI-resolved DataSource in JDBCQuotaStore.close()

groldan:
JNDI lookups return a DataSource owned by the JNDI provider; consumers must not close it, since the same name can be (and typically is) shared with other parts of the application.

Add an ownsDataSource flag (default true to keep the previous behavior for direct setDataSource callers) that JDBCQuotaStoreFactory clears whenever the configuration uses a JNDISource.


aaime (approved):
LGTM

Issue #1527, opened 2026-05-13
JDBCQuotaStore hits CannotSerializeTransactionException in clustered deployments


groldan:
When more than one JVM runs `DiskQuotaMonitor` against the same database (any clustered GeoServer deployment), concurrent updates to the same `TILESET` or `TILEPAGE` row deterministically abort with PostgreSQL's `could not serialize access due to concurrent update`:

```
org.springframework.dao.CannotSerializeTransactionException:
  PreparedStatementCallback;
  SQL [UPDATE TILEPAGE SET FILL_FACTOR = ? WHERE KEY = ?];
  ERROR: could not serialize access due to concurrent update
```

There is no retry, so the operation just fails. Easy to trigger: set a global quota lower than the cached size on two nodes - both `LayerQuotaEnforcementTask` instances try the same `UPDATE TILEPAGE SET FILL_FACTOR = 0 ...`.

### Why it matters

The wire-up for clustered GeoServer is essentially complete in the rest of the stack; the JDBC quota store is the missing piece. Today every write surface in `JDBCQuotaStore` is vulnerable when two JVMs touch the same row. The hottest case is the global `___GLOBAL_QUOTA___` row in TILESET, which **every tile-writing node updates on every batched flush**. With N replicas, that row is the cluster-wide hotspot.

### Root cause

`JDBCQuotaStore.java`:

```java
this.tt.setIsolationLevel(TransactionTemplate.ISOLATION_SERIALIZABLE);
```

SERIALIZABLE was a defensive choice for single-JVM correctness. There is no retry on serialization failure. Two hot paths additionally use a tolerant SELECT-then-INSERT-or-UPDATE loop in Java rather than an atomic upsert:

- `upsertTilePageFillFactor()`
- `upsertTilePageHitAccessTime()`

Both loop up to 100 times; under `SERIALIZABLE` the read + write within a single transaction triggers Postgres [SSI's predicate locking](https://wiki.postgresql.org/wiki/Serializable#Predicate_Locking) and aborts when two nodes touch the same `TILEPAGE` row.

The TILESET quota update is already an in-place SQL increment (`SQLDialect.java:326`: `UPDATE TILESET SET BYTES = BYTES + (:bytes) ...`), so it's safe at READ COMMITTED today; only the isolation level needs to change for that path.

### Proposed fix

> #### ~~Part 1 (the fix, all dialects): drop SERIALIZABLE~~
> 
[...]

PR #1528, opened 2026-05-14
diskquota/jdbc: Testcontainers-backed integration tests + CI workflow

groldan:
Add Testcontainers-driven integration tests for the JDBC DiskQuota store, so the PostgreSQL and Oracle dialects are actually exercised in CI - until now both were skipped whenever the `~/.geowebcache/<dialect>.properties`
fixture was absent.

The fixture-based `PostgreSQLQuotaStoreTest` / `OracleQuotaStoreTest` stays untouched.

## How to run locally

```shell
mvn -f geowebcache/pom.xml -pl :gwc-diskquota-jdbc -am \
    -Ponline -DskipTests=true -DskipITs=false verify
```

Drop `-DskipTests=true` to also run surefire (HSQL + unit tests).

## Known gap: Oracle IT is @Ignored

Running the Oracle IT against `gvenzl/oracle-xe:21-slim-faststart` expose a longstanding, dialect-level gap (not introduced here): any SERIALIZABLE transaction in `JDBCQuotaStore` whose first read goes through the freshly indexed TILEPAGE / TILESET fails with `ORA-08176: consistent read failure; rollback data not available`. As per Oracle's [own diagnostic](https://docs.oracle.com/error-help/db/ora-08176/), this is the documented behavior when a consistent-read snapshot must reach an operation that produces no undo, like `CREATE INDEX`.

The quota store creates 4 indexes on TILEPAGE at startup, so the very first SERIALIZABLE read of a fresh schema hits it.

Oracle's recommended remedy is exactly the retry-on-`CannotSerializeTransactionException` layer proposed in #1527 to address PostgreSQL's clustered SSI aborts. Once that lands, the Oracle IT will succeed on retry (unless other issues are found) and the `@Ignore` can be removed; the IT class javadoc explains the symptom and root cause in detail for the next reader.

We keep the Oracle IT class (rather than deleting it) so the container plumbing stays exercised up to `@ClassRule` startup, and there is a working scaffold ready for the retry-layer work.

### DIFF
--- a/.github/workflows/diskquota-jdbc-integration.yml
+++ b/.github/workflows/diskquota-jdbc-integration.yml
@@ -1,0 +1,43 @@
+name: DiskQuota JDBC Integration
+
+on:
+  push:
+    tags:
+      - "**"
+  pull_request:
+    paths:
+      - ".github/workflows/diskquota-jdbc-integration.yml"
+      - "pom.xml"
+      - "geowebcache/pom.xml"
+      - "geowebcache/core/**"
+      - "geowebcache/diskquota/**"
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  testcontainers:
+    name: DiskQuota JDBC Testcontainers (Postgres + Oracle XE)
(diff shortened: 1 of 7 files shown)


aaime:
Looks good, thanks!

PR #1530, opened 2026-05-15
diskquota/jdbc: retry SERIALIZABLE aborts (#1527)

### LINKED ISSUE #1527: JDBCQuotaStore hits CannotSerializeTransactionException in clustered deployments

groldan:
Closes #1527.
Requires #1528 #1529

`JDBCQuotaStore` runs every transaction at `SERIALIZABLE` isolation to defend the quota ledger against lost updates, but it never implemented the retry that `SERIALIZABLE` is defined in terms of. When two writers contended on the same
`TILESET` or `TILEPAGE` row  (e.g. several consumer threads in one JVM under load, or any clustered GeoServer deployment) the database aborts one of the transactions and Spring throws `DataAccessException` (a `RuntimeException`). The exception propagates to `QueuedQuotaUpdatesConsumer.call()`, which catches the`RuntimeException`, logs at `FINE`, and continues; the aggregated batch of pending quota updates was silently drop.

That silent drop is itself the lost-update path `SERIALIZABLE` was meant to prevent. Over time the in-DB ledger drifts out of sync with what's on disk the original symptom that motivated isolating quota writes in the first place, which might be hard to diagnose because if it happens sporadically, a large cache disk size (e.g. computed with `du -sh`) will not deviate enough from the computed disk quota enough to draw attention, and the FINE log level would just hide it.

## Fix

Wrap every `tt.execute(...)` call in `JDBCQuotaStore` with a bounded-retry helper. Default policy: 10 attempts, 10 ms -> 500 ms exponential backoff with full jitter, `FINE`-level logging on each retry and `WARNING` only when the cap is exhausted. The retry predicate matches three families of abort signals:

| Family | What it covers |
|---|---|
| `PessimisticLockingFailureException` (Spring) | Postgres SSI aborts (`SQLSTATE 40001`, translated by Spring to `CannotAcquireLockException`) |
| `SQLException` with `SQLState` class `"40"` | HSQL's `SQLTransactionRollbackException`, which Spring translates to a bare `ConcurrencyFailureException` rather than to one of the pessimistic-locking subclasses |
| `SQLException` with Oracle vendor code `8176` or `8177` | ORA-08176 (`consistent read failure`) and ORA-08177 (`can't serialize access`). Spring leaves 08176 uncategorised and routes 08177 to the deprecated `CannotSerializeTransactionException` (a sibling of `PessimisticLockingFailureException`, not a subclass), so vendor-code matching keeps the dialect-specific knowledge local to this predicate |

The `SERIALIZABLE` isolation level is **not** touched. The maintainer added it deliberately to defend the ledger; this PR is the missing application-level retry that brings the code in line with how `SERIALIZABLE` is meant to be used.

### Two important refinements once the retry loop was in place

- **Skip retry when nested inside an active transaction.** Spring's `PROPAGATION_REQUIRED` reuses the outer transaction, so an inner `executeWithRetry` (e.g. `createLayerInternal` called from `initialize`) that loops can only re-fail against the same stale snapshot. The helper now short-circuits with `TransactionSynchronizationManager.isActualTransactionActive()` and lets aborts bubble up to the outermost retry, which starts a fresh transaction. On Oracle this cut full-suite cap-exhaustion `WARNING`s from 22 to 0 in `OracleQuotaStoreIT`/`OracleQuotaStoreConcurrencyIT`.
- **Run DDL outside the wrapping transaction in `initialize()`.** Oracle auto-commits DDL and leaves the post-DDL SCN bookkeeping in a state where the first `SERIALIZABLE` read across recently-created indexes aborts with ORA-08176; running schema setup before the wrapped block lets the snapshot be taken cleanly past that auto-commit. HSQLDB also forces an implicit commit immediately before and after each DDL statement but isn't affected by the snapshot-read failure mode; Postgres, in contrast, has fully transactional DDL. The refactor is portably safe either way,`initialize()` is run-once startup work, and narrows the wrapped block to just the layer-reconciliation reads/writes that actually need retry.

### Oracle-specific completeness

Oracle cannot support `ON UPDATE CASCADE`, so the existing FK migration path (introduced in #1526 for the other dialects) is overridden:

- The `TILEPAGE -> TILESET` FK is declared `DEFERRABLE INITIALLY DEFERRED` on fresh installs and migrated to that shape on upgrades, removing the snapshot read against `TILESET` that fires on every plain `INSERT INTO TILEPAGE` (the main remaining ORA-08176 trigger).
- `getRenameLayerStatement` is rewritten as a PL/SQL anonymous block that rewrites both `TILESET.KEY` and `TILEPAGE.TILESET_ID` inside a single transaction. The deferred FK is verified once at commit with both updates already in place. This fixes the two `OracleQuotaStoreIT.testRenameLayer{,2}` cases that #1526's stricter assertions left failing.

The `SQLDialect` migrate scaffolding from #1526 is refactored to expose two protected hooks: `tilepageFkIsMigrated(rs)` and `tilepageFkAddSql(table, prefix)`. This way Oracle's overrides without duplicating the scan/race-recovery loop.

## Tests

A new `AbstractJDBCQuotaStoreConcurrencyTest` holds the shared scenarios. Three concrete subclasses run them against the dialects we support:

| Class | Phase | Engine |
|---|---|---|
| `HSQLQuotaStoreConcurrencyTest` | surefire | In-memory HSQL |
| `PostgreSQLQuotaStoreConcurrencyIT` | failsafe | Postgres testcontainer |
| `OracleQuotaStoreConcurrencyIT` | failsafe | Oracle XE testcontainer |

`JDBCQuotaStoreRetryTest` (offline) covers the retry helper itself: cap exhaustion, interrupt mid-backoff, and non-concurrency exceptions skipping the retry path entirely. The existing `JDBCQuotaStoreTest` suite passes on all dialects.

## Verification

Running `mvn -Ponline verify` on the diskquota/jdbc module:

```
Surefire:  90 run, 0 failures, 38 skipped (fixture-based PG/Oracle without ~/.geowebcache/*.properties)
Failsafe:  45 run, 0 failures
  PostgreSQLQuotaStoreIT                 19/19
  OracleQuotaStoreIT                     19/19    (~46 s)
  PostgreSQLForeignKeyMigrationIT         3/3
  PostgreSQLQuotaStoreConcurrencyIT       2/2
  OracleQuotaStoreConcurrencyIT           2/2     (~33 s)
  HSQLQuotaStoreConcurrencyTest           2/2     (surefire)

Cap-exhaustion WARNs in test logs: 0
ORA-08176 occurrences in logs:     0
ORA-08177 occurrences in logs:     0
```

For comparison, the same Oracle IT suite took ~120 s and produced 66 ORA-08176 stack traces in the logs before this PR.

## Out of scope

- **Dropping `SERIALIZABLE`.** Deliberately kept. The retry is the missing piece, not the isolation model.
- **A Postgres-native `ON CONFLICT` upsert in `PostgreSQLDialect`.** Could reduce the conflict rate but does not replace the retry layer and haven't checked it'd work for all cases.

### DIFF
--- a/geowebcache/diskquota/jdbc/src/main/java/org/geowebcache/diskquota/jdbc/OracleDialect.java
+++ b/geowebcache/diskquota/jdbc/src/main/java/org/geowebcache/diskquota/jdbc/OracleDialect.java
@@ -13,6 +13,9 @@
  */
 package org.geowebcache.diskquota.jdbc;
 
+import java.sql.DatabaseMetaData;
+import java.sql.ResultSet;
+import java.sql.SQLException;
 import java.util.List;
 
 /**
@@ -60,7 +63,8 @@ public class OracleDialect extends SQLDialect {
                         """
                         CREATE TABLE ${schema}TILEPAGE (
                          KEY VARCHAR(%d) PRIMARY KEY,
-                         TILESET_ID VARCHAR(%d) REFERENCES ${schema}TILESET(KEY) ON DELETE CASCADE,
+                         TILESET_ID VARCHAR(%d) REFERENCES ${schema}TILESET(KEY) ON DELETE CASCADE
+                           DEFERRABLE INITIALLY DEFERRED,
                          PAGE_Z SMALLINT,
                          PAGE_X INTEGER,
                          PAGE_Y INTEGER,
@@ -84,37 +88,46 @@ public class OracleDialect extends SQLDialect {
     }
 
     /**
-     * No-op: Oracle does not support {@code ON UPDATE CASCADE} on foreign keys, so there is nothing portable to
-     * migrate. Companion to {@link #getRenameLayerStatement(String, String, String)}, which preserves the legacy
-     * LAYER_NAME-only behavior on this dialect.
+     * Oracle does not support {@code ON UPDATE CASCADE}, so the FK is migrated to {@code DEFERRABLE INITIALLY DEFERRED}
+     * instead. Deferring the check to commit time also drops the per-INSERT snapshot read on TILESET that triggers
+     * ORA-08176 under SERIALIZABLE.
      */
     @Override
-    public void migrateForeignKeys(String schema, SimpleJdbcTemplate template) {
-        // intentional no-op
+    protected boolean tilepageFkIsMigrated(ResultSet rs) throws SQLException {
+        return rs.getShort("DEFERRABILITY") == DatabaseMetaData.importedKeyInitiallyDeferred;
+    }
+
+    @Override
+    protected String tilepageFkAddSql(String prefixedTilepageName, String prefix) {
+        return """
+                ALTER TABLE %s ADD FOREIGN KEY (TILESET_ID)
+                REFERENCES %sTILESET(KEY)
+                ON DELETE CASCADE
+                DEFERRABLE INITIALLY DEFERRED
+                """
+                .formatted(prefixedTilepageName, prefix);
     }
 
     /**
-     * Oracle does not support {@code ON UPDATE CASCADE} on foreign keys, so the {@code TILEPAGE.TILESET_ID -> TILESET
-     * .KEY} FK declared above only cascades on delete. As a result this dialect cannot safely rewrite {@code TILESET
-     * .KEY} during a rename without first dealing with the dangling {@code TILEPAGE} rows.
-     *
-     * <p>For now Oracle keeps the legacy behavior of only updating {@code LAYER_NAME}; lookups by id against the
-     * renamed layer will continue to miss the row and cause {@code getOrCreateTileSet} to insert duplicates. Fixing
-     * this on Oracle (e.g. via {@code DEFERRABLE INITIALLY DEFERRED} constraints, or by disabling the FK around the
-     * rename) is tracked separately.
+     * PL/SQL anonymous block that rewrites TILESET.KEY and TILEPAGE.TILESET_ID together; the deferred FK is checked
+     * once at commit with both updates in place. Oracle has no {@code ON UPDATE CASCADE}, hence the manual rewrite, and
+     * no SQL-standard {@code SUBSTRING ... FROM POSITION(...)}, hence {@code SUBSTR}/{@code INSTR}.
      */
     @Override
     public String getRenameLayerStatement(String schema, String oldLayerName, String newLayerName) {
-        StringBuilder sb = new StringBuilder("UPDATE ");
-        if (schema != null) {
-            sb.append(schema).append(".");
-        }
-        sb.append("TILESET SET LAYER_NAME = :")
-                .append(newLayerName)
-                .append(" WHERE LAYER_NAME = :")
-                .append(oldLayerName);
-
-        return sb.toString();
+        String prefix = schema == null ? "" : schema + ".";
+        return """
+                BEGIN
+                  UPDATE %sTILESET
+                    SET KEY = :%s || SUBSTR(KEY, INSTR(KEY, '#')),
+                        LAYER_NAME = :%s
+                    WHERE LAYER_NAME = :%s;
+                  UPDATE %sTILEPAGE
+                    SET TILESET_ID = :%s || SUBSTR(TILESET_ID, INSTR(TILESET_ID, '#'))
+                    WHERE INSTR(TILESET_ID, :%s || '#') = 1;
+                END;
+                """
+                .formatted(prefix, newLayerName, newLayerName, oldLayerName, prefix, newLayerName, oldLayerName);
     }
 
     @Override
--- a/geowebcache/diskquota/jdbc/src/main/java/org/geowebcache/diskquota/jdbc/SQLDialect.java
+++ b/geowebcache/diskquota/jdbc/src/main/java/org/geowebcache/diskquota/jdbc/SQLDialect.java
@@ -172,26 +172,19 @@ public class SQLDialect {
         try (ResultSet rs = dbmd.getImportedKeys(null, schema, tilepageName)) {
             while (rs.next()) {
                 String fkName = rs.getString("FK_NAME");
-                if (!isTilesetCascadeCandidate(rs, fkName)) {
+                if (!isTilepageTilesetFkRow(rs, fkName) || tilepageFkIsMigrated(rs)) {
                     continue;
                 }
                 String drop = "ALTER TABLE %s DROP CONSTRAINT %s".formatted(prefixedTilepageName, fkName);
-                String add =
-                        """
-                        ALTER TABLE %s ADD FOREIGN KEY (TILESET_ID)
-                        REFERENCES %sTILESET(KEY)
-                        ON UPDATE CASCADE ON DELETE CASCADE
-                        """
-                                .formatted(prefixedTilepageName, prefix);
+                String add = tilepageFkAddSql(prefixedTilepageName, prefix);
 
-                LOG.info(() -> "Upgrading TILEPAGE.TILESET_ID foreign key to ON UPDATE CASCADE (was constraint %s)"
-                        .formatted(fkName));
+                LOG.info(() -> "Migrating TILEPAGE.TILESET_ID foreign key (was constraint %s)".formatted(fkName));
                 JdbcOperations jdbcOperations = template.getJdbcOperations();
                 try {
                     jdbcOperations.execute(drop);
                     jdbcOperations.execute(add);
                 } catch (DataAccessException raceLikely) {
-                    if (isTilepageFkAlreadyCascade(dbmd, schema, tilepageName)) {
+                    if (isTilepageFkAlreadyMigrated(dbmd, schema, tilepageName)) {
                         LOG.fine(() -> "TILEPAGE FK was migrated concurrently by another instance "
                                 + "while this instance was trying to drop %s; accepting concurrent migration"
                                         .formatted(fkName));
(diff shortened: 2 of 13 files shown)


aaime (changes requested):
Overall looks good, two feedbacks:
* I think the foreign key fix got merged into the PR by accident?
* Some of the explanations javadocs are very long-winded, while reviewing I am repeatedly tempted to jump to the code rather than keep on reading. I would try to summarize a bit and only comment on bits that are less than obvious.

groldan:
hi, thanks for reviewing.

> I think the foreign key fix got merged into the PR by accident?

no, I jus rebased it after #1529 was merged

> Some of the explanations javadocs are very long-winded

hear you, shriking... brb

geowebcache/diskquota/jdbc/src/main/java/org/geowebcache/diskquota/jdbc/OracleDialect.java
```diff
@@ -84,37 +88,46 @@ protected void addEmtpyTableReference(StringBuilder sb) {
...
+                  UPDATE %sTILESET
+                    SET KEY = :%s || SUBSTR(KEY, INSTR(KEY, '#')),
+                        LAYER_NAME = :%s
+                    WHERE LAYER_NAME = :%s;
+                  UPDATE %sTILEPAGE
+                    SET TILESET_ID = :%s || SUBSTR(TILESET_ID, INSTR(TILESET_ID, '#'))
+                    WHERE TILESET_ID LIKE :%s || '#%%';
+                END;
+                """
+                .formatted(prefix, newLayerName, newLayerName, oldLayerName, prefix, newLayerName, oldLayerName);
```
aaime, line 130: Hum... what happens when a layer name contains the underscore? It's not escaped, and in a SQL like it's the single character wildcard

geowebcache/diskquota/jdbc/src/main/java/org/geowebcache/diskquota/jdbc/SQLDialect.java
```diff
@@ -172,26 +172,19 @@ private Void upgradeTilepageForeignKey(DatabaseMetaData dbmd, String schema, Sim
...
+                LOG.info(() -> "Migrating TILEPAGE.TILESET_ID foreign key (was constraint %s)".formatted(fkName));
                 JdbcOperations jdbcOperations = template.getJdbcOperations();
                 try {
                     jdbcOperations.execute(drop);
                     jdbcOperations.execute(add);
                 } catch (DataAccessException raceLikely) {
-                    if (isTilepageFkAlreadyCascade(dbmd, schema, tilepageName)) {
+                    if (awaitConcurrentMigration(dbmd, schema, tilepageName)) {
                         LOG.fine(() -> "TILEPAGE FK was migrated concurrently by another instance "
                                 + "while this instance was trying to drop %s; accepting concurrent migration"
```
aaime: This code is problematic, even if for a short window of time, on Oracle, because DDL is not transactional. So the drop can run, complete succesfully, and then the Oracle node can go down for whatever reason (restart, network) and never issue the "add", while leaving the drop in place. 
Suggestion: run the drop by itself, and then follow up with a separate try/catch handling the add, and the retry logic should consider that the starting state is "no more foreign key"

aaime (commented):
This one was pretty hard to review. I migth have found a couple of issues, see below.

geowebcache/diskquota/jdbc/src/main/java/org/geowebcache/diskquota/jdbc/OracleDialect.java
groldan: Fixed: the `TILEPAGE` update matches the prefix with `INSTR()` instead of `LIKE` (hence `_`, `%`, `.`, etc.) in a layer name are treated literally
```
    WHERE INSTR(TILESET_ID, :%s || '#') = 1
```

geowebcache/diskquota/jdbc/src/main/java/org/geowebcache/diskquota/jdbc/SQLDialect.java
groldan: Good catch. Fixed in e6d6c84027f4314d465d892ee85bd6f2c2b7aff3. The migration no longer assumes drop+add is atomic.

- Now DROP and ADD are separate committed statements. `upgradeTilepageForeignKey` converges on the migrated FK from any state: already migrated, legacy present, or **no FK at all** (e..g interrupted prior migration or Oracle down) re-adds it instead of skipping the table.
- On a failed step it polls the live FK and accepts a migration completed by a peer nsteat of rethrowing.

Regression test `migrateRestoresForeignKeyDroppedByInterruptedMigration` drops the FK and asserts the next `migrateForeignKeys` re-adds it.

Also moved the orchestration out of `SQLDialect` into a dedicated `TilepageForeignKeyMigrator`.

PR #1533, opened 2026-05-22
using GT utility classes for XML parsing/producing

pmauduit:
See related PRs on GT/GS for the motivation:

* https://github.com/geotools/geotools/pull/5639
* https://github.com/geoserver/geoserver/pull/9502

Making use of GT utlity class to get XML-parsing related objects so that they are obtained from a centralized source.

### DIFF
--- a/geowebcache/core/pom.xml
+++ b/geowebcache/core/pom.xml
@@ -24,6 +24,10 @@
       <artifactId>gt-coverage</artifactId>
     </dependency>
     <dependency>
+      <groupId>org.geotools</groupId>
+      <artifactId>gt-xml</artifactId>
+    </dependency>
+    <dependency>
       <groupId>org.apache.commons</groupId>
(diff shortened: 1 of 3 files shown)


jodygarnett:
Hi @pmauduit I would like to merge to `default-entity-resolver-check` so that the upstream GeoTools integration tests can run.

With that in mind I will cherry pick your two commits over to that branch.

Q: Do you have access to geowebcache repo to work on public branch?

pmauduit:
I'm missing 2 places in the code where I could not use GT's XMLUtils, I don't have access to GWC, so the current pr has been opened from my fork.

jodygarnett:
This PR is supersseded by https://github.com/GeoWebCache/geowebcache/pull/1534 which makes use of public `default-entity-resolver-check` branch which is subject to integration tests.

PR #1430, opened 2025-09-06
fix-1429 maxPoolSize and maxThreadSize should be configurable

draganpodvezanec-verox:
closes #1429 

The seeder thread pool configuration now uses environment variables with fallback to the original default values:

Core pool size: ${GWC_SEEDER_CORE_POOL_SIZE:16}
Uses the GWC_SEEDER_CORE_POOL_SIZE environment variable, defaulting to 16 if not set

Maximum pool size: ${GWC_SEEDER_MAX_POOL_SIZE:32}
Uses the GWC_SEEDER_MAX_POOL_SIZE environment variable, defaulting to 32 if not set

PR #1535, opened 2026-05-28
Make seeder thread pool sizes configurable (fixes #1429)

### LINKED ISSUE #1429: maxPoolSize and maxThreadSize should be configurable
After applying PR
https://github.com/GeoWebCache/geowebcache/pull/1427

...we can establish a much larger number of HTTP connections to the local Geoserver pool, and seed many layers in parallel. But there is a limitation on max seeder threads, which is hardcoded:

```
<!-- Thread pool for seeding -->
  <bean id="gwcSeederThreadPoolExec" 
    class="org.geowebcache.seed.SeederThreadPoolExecutor">
    <constructor-arg value="16"/><!-- Size of core pool -->
    <constructor-arg value="32"/><!-- Maximum size of pool -->
  </bean>
```

This should be possible to configure using the environment variable.

petersmythe:
## Summary

Makes the GeoWebCache seeder thread pool sizes configurable via `geowebcache.xml` (primary configuration) with environment variable overrides for Docker/cloud deployments, as requested in #1429.

Previously, the core pool size (16) and max pool size (32) were hardcoded in Spring XML with no way to override them without modifying the application context.

### Primary: geowebcache.xml

```xml
<gwcConfiguration>
  ...
  <seederCorePoolSize>24</seederCorePoolSize>
  <seederMaxPoolSize>64</seederMaxPoolSize>
  ...
</gwcConfiguration>
```

| Element | Description | Default |
|---------|-------------|---------|
| `seederCorePoolSize` | Core thread pool size for seeding tasks | 16 |
| `seederMaxPoolSize` | Maximum thread pool size for seeding tasks | 32 |

### Override: Environment variables

For Docker/cloud deployments, the XML configuration can be overridden:

| Override | Overrides |
|----------|-----------|
| `GWC_SEEDER_CORE_POOL_SIZE` | `seederCorePoolSize` |
| `GWC_SEEDER_MAX_POOL_SIZE` | `seederMaxPoolSize` |

**Precedence:** Environment variable / system property (`-D`) → `geowebcache.xml` → hardcoded default (16/32).

## Changes

- `ServerConfiguration`: Added `getSeederCorePoolSize()` / `getSeederMaxPoolSize()` default interface methods.
- `GeoWebCacheConfiguration`: Added `seederCorePoolSize` / `seederMaxPoolSize` fields (POJO for geowebcache.xml).
- `geowebcache.xsd`: Added `<seederCorePoolSize>` and `<seederMaxPoolSize>` schema elements.
- `geowebcache.xml` (example): Added commented-out example.
- `geowebcache-core-context.xml`: Bean wiring now reads from `gwcXmlConfig` via SpEL, falling back to 16/32.
- `SeederThreadPoolExecutor`: Added `resolvePoolSize()` for env var override with validation (invalid values fall back to defaults, core > max auto-adjusts).
- New `SeederThreadPoolExecutorTest` with 9 test cases.
- Documentation updated in `production/index.rst`.

## Backward Compatibility

- The Spring XML bean definitions now read from the configuration object but fall back to 16/32 if not set — existing deployments without `seederCorePoolSize`/`seederMaxPoolSize` in their `geowebcache.xml` continue to work unchanged.
- `SeederThreadLocalTransferExecutor` (GeoServer subclass) works without modification.
- No changes to the constructor signature.

## Testing

- 9 unit tests covering: defaults, system property override, partial override, invalid values (non-numeric, negative, zero), whitespace handling, trimming, and corePoolSize > maxPoolSize adjustment.
- Manually tested in Docker with GeoServer 3.0.x nightly — confirmed 24 threads active matching `GWC_SEEDER_CORE_POOL_SIZE=24` via `jcmd Thread.print`.

## Supersedes

This replaces the approach in #1430 which used Spring property placeholders in the XML. That approach had test failures and didn't meet the requirement of being configurable outside of application contexts.

Closes #1429

### DIFF
--- a/geowebcache/core/src/main/java/org/geowebcache/seed/SeederThreadPoolExecutor.java
+++ b/geowebcache/core/src/main/java/org/geowebcache/seed/SeederThreadPoolExecutor.java
@@ -61,12 +61,9 @@ public class SeederThreadPoolExecutor extends ThreadPoolExecutor implements Disp
         int core = resolvePoolSize(GWC_SEEDER_CORE_POOL_SIZE, defaultCore);
         int max = resolvePoolSize(GWC_SEEDER_MAX_POOL_SIZE, defaultMax);
         if (core > max) {
-            log.warning(GWC_SEEDER_CORE_POOL_SIZE
-                    + " ("
+            log.warning("Configured corePoolSize ("
                     + core
-                    + ") is greater than "
-                    + GWC_SEEDER_MAX_POOL_SIZE
-                    + " ("
+                    + ") is greater than maxPoolSize ("
                     + max
                     + "), adjusting maxPoolSize to match corePoolSize");
             max = core;
--- a/geowebcache/core/src/main/java/org/geowebcache/seed/SeederThreadPoolExecutor.java
+++ b/geowebcache/core/src/main/java/org/geowebcache/seed/SeederThreadPoolExecutor.java
@@ -19,8 +19,10 @@ import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
 import java.util.logging.Logger;
 import org.geotools.util.logging.Logging;
+import org.geowebcache.config.ServerConfiguration;
 import org.springframework.beans.factory.DisposableBean;
 import org.springframework.scheduling.concurrent.CustomizableThreadFactory;
+import org.springframework.util.StringUtils;
 
 public class SeederThreadPoolExecutor extends ThreadPoolExecutor implements DisposableBean {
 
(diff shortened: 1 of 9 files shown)


aaime:
GeoWebCache has a configuration file, it's geowebcache.xml, new configurations should go in there... an env variable override makes some sense too, but as an override, not straight configuration. Most of the GWC configuration is in that XML file, we should not create a parallel configuration system just because it's convenient to code.

petersmythe:
> GeoWebCache has a configuration file, it's geowebcache.xml, new configurations should go in there... an env variable override makes some sense too, but as an override, not straight configuration. Most of the GWC configuration is in that XML file, we should not create a parallel configuration system just because it's convenient to code.

Thank you, of course it should. I will rework.

petersmythe:
Ready and tested again.  Thanks for the feedback.

geowebcache/core/src/main/java/org/geowebcache/seed/SeederThreadPoolExecutor.java
```diff
@@ -28,8 +28,87 @@ public class SeederThreadPoolExecutor extends ThreadPoolExecutor implements Disp
...
+     * then as an OS environment variable. Falls back to the provided default if neither is set or the value is not a
+     * valid positive integer.
+     *
+     * @param propertyName the system property / environment variable name to look up
+     * @param defaultValue the fallback value if the property is not set or invalid
+     * @return the resolved pool size
+     */
+    static int resolvePoolSize(String propertyName, int defaultValue) {
+        String value = System.getProperty(propertyName);
+        if (value == null || value.isBlank()) {
```
pmauduit: `org.springframework.util.StringUtils.hasText(value)` if you want to avoid the null check

geowebcache/core/src/main/java/org/geowebcache/seed/SeederThreadPoolExecutor.java
```diff
@@ -28,8 +28,87 @@ public class SeederThreadPoolExecutor extends ThreadPoolExecutor implements Disp
...
+     * integer, the {@code maxPoolSize} constructor argument is used as the default.
+     */
+    public static final String GWC_SEEDER_MAX_POOL_SIZE = "GWC_SEEDER_MAX_POOL_SIZE";
+
     public SeederThreadPoolExecutor(int corePoolSize, int maxPoolSize) {
-        super(corePoolSize, maxPoolSize, 60, TimeUnit.SECONDS, new LinkedBlockingQueue<>(), tf);
+        this(resolveAndValidateSizes(corePoolSize, maxPoolSize));
+    }
+
+    private SeederThreadPoolExecutor(int[] sizes) {
```
pmauduit, line 71: I find the array as argument a bit confusing to read, but since the method is private, it should be ok.

geowebcache/core/src/test/java/org/geowebcache/seed/SeederThreadPoolExecutorTest.java
```diff
@@ -0,0 +1,157 @@
+/**
+ * This program is free software: you can redistribute it and/or modify it under the terms of the GNU Lesser General
```
pmauduit, line 2: missing copyright / year ?

geowebcache/web/src/main/webapp/WEB-INF/geowebcache-core-context.xml
```diff
@@ -180,8 +180,8 @@
   <!-- Thread pool for seeding -->
   <bean id="gwcSeederThreadPoolExec" 
     class="org.geowebcache.seed.SeederThreadPoolExecutor">
-    <constructor-arg value="16"/><!-- Size of core pool -->
-    <constructor-arg value="32"/><!-- Maximum size of pool -->
+    <constructor-arg value="#{gwcXmlConfig.seederCorePoolSize != null ? gwcXmlConfig.seederCorePoolSize : 16}"/><!-- Size of core pool -->
+    <constructor-arg value="#{gwcXmlConfig.seederMaxPoolSize != null ? gwcXmlConfig.seederMaxPoolSize : 32}"/><!-- Maximum size of pool -->
```
pmauduit: won't it be more "spring" to have the gwcXmlConfig bean as a dependency here and call the getters into the constructor ?

pmauduit (changes requested):

petersmythe:
Thank you @pmauduit 

Addressed review feedback - refactored to bean injection (<constructor-arg ref="gwcXmlConfig"/>) instead of SpEL, plus trivial fixes (StringUtils.hasText, copyright year, test rename).

geowebcache/web/src/main/webapp/WEB-INF/geowebcache-core-context.xml
petersmythe: Of course.  Sorry

PR #1536, opened 2026-05-30
Improve dispatcher error logging for client errors

petersmythe:
Catch IllegalArgumentException separately in GeoWebCacheDispatcher and log at WARNING level with message only (no stack trace).

These are client errors (e.g. requesting an unsupported format like 'image/png is not a supported format for layer_name') that do not warrant a full stack trace at SEVERE level. The client still receives a proper 400 response with the error message.

### DIFF
--- a/geowebcache/core/src/main/java/org/geowebcache/GeoWebCacheDispatcher.java
+++ b/geowebcache/core/src/main/java/org/geowebcache/GeoWebCacheDispatcher.java
@@ -316,6 +316,9 @@ public class GeoWebCacheDispatcher extends AbstractController {
                     CacheResult.OTHER,
                     runtimeStats);
             LOG.warning(e.getMessage());
+        } catch (IllegalArgumentException e) {
+            ResponseUtils.writeErrorPage(response, 400, e.getMessage(), runtimeStats);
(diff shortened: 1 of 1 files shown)


petersmythe:
Waiting on https://github.com/geoserver/geoserver/pull/9538 to fix the broken tests

petersmythe:
Ignoring broken downstream (GeoServer) tests

aaime:
The premise that IllegalArgumentException represents a client error is incorrect. It is a general-purpose Java runtime exception thrown throughout the codebase for a wide range of conditions, including internal programming errors. Catching it at the dispatcher level and returning a 400 will silently misclassify genuine bugs as client mistakes, and dropping the stack trace means those bugs become undiagnosable in production.

Additionally, a behavior change of this kind should not be self-merged after 20 hours without review. You probably thought it was a harmless, small change, which is probably even worse, it shows that you have no understanding of the code change consequences. 

Direct merges without review are for emergency usage only, and should be performed only by someone with a clear understanding of what they are doing.

petersmythe:
I am duly chastised. I am sorry.

petersmythe:
Superseded by https://github.com/geoserver/geoserver/pull/9552

PR #1541, opened 2026-06-04
Make use of XMLUtils.supportDTD to maintain WMS DTD support

jodygarnett:
Companion to:

* https://github.com/geotools/geotools/pull/5691
* https://github.com/geoserver/geoserver/pull/9585

Restrictions on XMLUtils required DTD to be enabled when required.


jodygarnett:
Restart builds to check against GeoTools SNAPSHOTs

PR #1545, opened 2026-06-18
Allow parameter filters to be private and not exposed in capabilities OL preview and the like

aaime:
In preparation for https://github.com/geoserver/geoserver/wiki/GSIP-241, which is adding on the fly a couple of of such parameters (the security rules summary and the security tags)


dromagnoli (approved):
Looks good to me

